PR #5, opened 2019-07-12
Allow users to pass a single dir path to render

celebdor:
This makes it so that all the *.tmpl files in the directory get rendered


tomassedovic:
FWIW I've tried this and it works. +1 from me

PR #7, opened 2019-07-16
haproxy monitor addition

celebdor:
HAProxy needs a sidecar container that monitors it. This patch adds a new
monitor command that implements that functionality.


pkg/monitor/iptables.go
```diff
@@ -0,0 +1,61 @@
...
+	if exists, _ := ipt.Exists(table, chain, ruleSpec...); exists {
+		log.WithFields(logrus.Fields{
+			"spec": strings.Join(ruleSpec, " "),
+		}).Info("Removing existing nat PREROUTING rule")
+		return ipt.Delete(table, chain, ruleSpec...)
+	}
+	return nil
+}
+
+func ensureHAProxyPreRoutingRule(apiVip string, apiPort, lbPort uint16) error {
```
yboaron: I think it will be very useful (debugging)to add log messages in iptable rules functions

celebdor: Got it. I noticed that I was ignoring the err from this function in the caller. I'm adding a log message there.

PR #14, opened 2019-08-01
Monitor - delete iptable rule if API not reachable through Haproxy

yboaron:
If Haproxy went down for some reason, the Iptable rule that
redirects API traffic to Haproxy LB, should be removed.

### DIFF
--- a/pkg/monitor/monitor.go
+++ b/pkg/monitor/monitor.go
@@ -93,6 +93,7 @@ func GetLBConfig(domain string, apiPort, lbPort, statPort uint16) (ApiLBConfig, 
 
 func Monitor(clusterName, clusterDomain, templatePath, cfgPath, apiVip string, apiPort, lbPort, statPort uint16, interval time.Duration) error {
 	var oldConfig, newConfig *ApiLBConfig
+	var k8sIsHealthy bool = false
 	signals := make(chan os.Signal, 1)
 	done := make(chan bool, 1)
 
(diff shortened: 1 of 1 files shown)


pkg/monitor/monitor.go
```diff
@@ -144,16 +146,21 @@ func Monitor(clusterName, clusterDomain, templatePath, cfgPath, apiVip string, a
 
 			ok, err := utils.IsKubernetesHealthy(lbPort)
 			if err == nil && ok {
-				log.Info("API reachable through HAProxy")
+				if k8sIsHealthy == false {
```
celebdor:
```suggestion
				if ! k8sIsHealthy {
```

pkg/monitor/monitor.go
```diff
@@ -144,16 +146,21 @@ func Monitor(clusterName, clusterDomain, templatePath, cfgPath, apiVip string, a
...
+					k8sIsHealthy = true
+				}
 				err := ensureHAProxyPreRoutingRule(apiVip, apiPort, lbPort)
 				if err != nil {
 					log.WithFields(logrus.Fields{"err": err}).Error("Failed to ensure HAProxy PREROUTING rule to direct traffic to the LB")
 				}
 			} else {
-				log.Info("API still not reachable through HAProxy")
+				cleanHAProxyPreRoutingRule(apiVip, apiPort, lbPort)
+				if k8sIsHealthy == true {
```
celebdor:
```suggestion
				if k8sIsHealthy {
```

PR #19, opened 2019-09-16
Add stabilization mechanism to monitor logic

yboaron:
Monitor checks periodically for changes in masters topology, K8S API health via Haproxy
and updates load balancer configuration correspondingly.
To make the monitoring process more robust, this commit adds stabilization mechanism to
the periodic checks, means that Monitor will update load balancer configuration
only after it detects a change for N consecutive intervals.

### DIFF
--- a/pkg/monitor/monitor.go
+++ b/pkg/monitor/monitor.go
@@ -17,6 +17,9 @@ import (
 )
 
 const haproxyMasterSock = "/var/run/haproxy/haproxy-master.sock"
+const cfgChangeThreshold uint8 = 3
+const k8sHealthThresholdOn uint8 = 3
+const k8sHealthThresholdOff uint8 = 2
 
 var log = logrus.New()
 
@@ -25,8 +28,12 @@ type RuntimeConfig struct {
 }
 
 func Monitor(clusterName, clusterDomain, templatePath, cfgPath, apiVip string, apiPort, lbPort, statPort uint16, interval time.Duration) error {
-	var oldConfig, newConfig *config.ApiLBConfig
-	var k8sIsHealthy bool = false
+	var appliedConfig, curConfig, prevConfig *config.ApiLBConfig
+	var K8sHealthSts bool = false
+	var oldK8sHealthSts bool
+	var k8sHealthChangeCtr uint8 = 0
+	var configChangeCtr uint8 = 0
+
 	signals := make(chan os.Signal, 1)
 	done := make(chan bool, 1)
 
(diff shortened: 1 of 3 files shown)


pkg/monitor/monitor.go
```diff
@@ -119,44 +126,54 @@ func Monitor(clusterName, clusterDomain, templatePath, cfgPath, apiVip string, a
...
-			} else {
-				cleanHAProxyPreRoutingRule(apiVip, apiPort, lbPort)
-				if k8sIsHealthy {
-					log.Info("API is not reachable through HAProxy")
-					k8sIsHealthy = false
-				}
+			prevConfig = &config
+
+			curK8sHealthkSts, err := utils.IsKubernetesHealthy(lbPort)
+			if err != nil {curK8sHealthkSts=false}
```
celebdor: what does `curK8sHealthkSts` stand for?

yboaron: The idea is to have stabilization for k8s health status (for both on and off), so  'curK8sHealthkSts' holds the value of single (latest) k8s health check and 'K8sHealthSts' holds the value after stabilization.

The only usage of 'curK8sHealthkSts' is a parameter to stabilization function, and is being calculated as follows:

'curK8sHealthkSts, err := utils.IsKubernetesHealthy(lbPort)
if err != nil {curK8sHealthkSts=false}
'
I thought at the beginning to pass the output of 'utils.IsKubernetesHealthy(lbPort)'  as a parameter to the stabilization function (and not define a local variable for that case), but I don't know how to handle the case where err != NIL (new to GOLANG)

If you think there's a way to avoid using a local variable for that purpose, I'll be glad to change it.

pkg/monitor/monitor.go
```diff
@@ -25,8 +29,12 @@ type RuntimeConfig struct {
 }
 
 func Monitor(clusterName, clusterDomain, templatePath, cfgPath, apiVip string, apiPort, lbPort, statPort uint16, interval time.Duration) error {
-	var oldConfig, newConfig *config.ApiLBConfig
-	var k8sIsHealthy bool = false
+	var appliedConfig, curConfig, prevConfig *ApiLBConfig
```
cybertron: Go vet is failing on this. I'm assuming it needs to be *config.ApiLBConfig like it was before.

pkg/monitor/monitor.go
cybertron: Is Healthk a typo?

pkg/monitor/monitor.go
```diff
@@ -55,44 +63,52 @@ func Monitor(clusterName, clusterDomain, templatePath, cfgPath, apiVip string, a
...
-				_, err = conn.Write([]byte("reload\n"))
-				if err != nil {
-					log.WithFields(logrus.Fields{
-						"socket": haproxyMasterSock,
-					}).Error("Failed to write reload to HAProxy master socket")
-					return err
-				}
+			curConfig = &config
+			if appliedConfig == nil || !cmp.Equal(*appliedConfig, *curConfig) {
+			    if prevConfig == nil || cmp.Equal(*prevConfig, *curConfig) {configChangeCtr++}else {configChangeCtr=1}
```
cybertron: This may be my python background speaking, but I'm not a big fan of putting if...else statements on one line.

And in fact, neither is go fmt. I'm not sure why the ci job didn't catch this, but if I run go fmt against this change locally I get a large number of changes. That should probably be cleaned up before it merges.

pkg/monitor/monitor.go
yboaron: Yep, 10x!

pkg/monitor/monitor.go
yboaron: Ohh,  I missed that on the rebase process - 10x!

PR #20, opened 2019-09-18
Add dynkeepalived to solve interface changes

celebdor:
Using runtimecfg render for keepalived has the issue that on 1.x
keepalived versions there's no dynamic interfaces and in case the
interface changes, keepalived will stop doing its job.

Thanks to dynkeepalived, it is now possible to have it monitor the
interface and dynamically send a reload to a unix socket that can be
used to trigger a keepalived reload in the keepalived container.

This is meant to be used from a side-car to the keepalived pod.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,4 @@ network_closure.sh
 # binary
 runtimecfg
 monitor
+dynkeepalived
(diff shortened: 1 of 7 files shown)


celebdor:
/hold

Needs proper regression testing on the monitor and the runtimecfg binaries.

celebdor:
Did manual testing of the binaries and made a successful deployment to rule out regressions.

celebdor:
Verified with success that the IP was moved to brext and the VIPs did as well. The downtime for reconfiguration of the interfaces was of about 90seconds.

Issue #21, opened 2019-09-23
FletcherChecksum8() may generate invalid VRID


mandre:
Sometimes, keepalived pod may end up in a crash loop with the following logs.

```
Starting Keepalived v1.3.5 (03/19,2017), git commit v1.3.5-6-g6fa32f2
Opening file '/etc/keepalived/keepalived.conf'.
Starting VRRP child process, pid=7
Registering Kernel netlink reflector
Registering Kernel netlink command channel
Registering gratuitous ARP shared channel
Opening file '/etc/keepalived/keepalived.conf'.
Truncating auth_pass to 8 characters
VRRP Error : VRID not valid - must be between 1 & 255. reconfigure !
Truncating auth_pass to 8 characters
VRRP_Instance(l3vvqqni-f21ca_DNS) the virtual id must be set!
Stopped
Keepalived_vrrp exited with permanent error CONFIG. Terminating
Stopping
Stopped Keepalived v1.3.5 (03/19,2017), git commit v1.3.5-6-g6fa32f2
```

It turns out, the `FletcherChecksum8()` function used to compute a unique VRID may return `0` with certain input such as "l3vvqqni-f21ca-dns.
`0` is not a valid VRID.

https://play.golang.org/p/V0i11g1XvsY

cybertron:
Hmm, apparently the keepalived docs lie: "arbitrary unique number from 0 to 255" (from the virtual_router_id section of https://www.keepalived.org/manpage.html).

Also, comparing our fletcher8 implementation to some online calculators I'm getting different results for a given input value. I think to make ours "correct", we need to do:

-               ckA = (ckA + inp[i]) & 0xf
-               ckB = (ckB + ckA) & 0xf
+               ckA = (ckA + inp[i]) % 0xf
+               ckB = (ckB + ckA) % 0xf

However, for our purposes I don't think it matters. Both implementations give us an 8 bit value that changes depending on the input.

mandre:
I believe this has been fixed with #44 
/close

Issue #22, opened 2019-09-23
Possible VRID collision?


mandre:
Not sure this is really the case (haven't observed it), the use of a checksum algo to compute the VRID could lead to collision -- especially on a small pool of 255 possible values.

Would it be possible to add a check in the code to ensure the VRIDs are different?

cybertron:
This is definitely a possibility, and it's even worse than just collisions between our three VIPs. It's possible a deployer could have keepalived running on their network already for other services and we might accidentally pick one of their existing VRIDs. Ideally we would check for ids already in use on the network when we select ours and have a consistent fallback mechanism to deal with collisions.

The check in the code should be pretty easy to write, but I'm not sure how looking for existing ids on the network would work. So far it hasn't been a big problem so we haven't prioritized solving it.

mandre:
This has been fixed with #23 
/close

PR #23, opened 2019-09-23
Tweak VRID calculation to avoid 0

cybertron:
This change modifies how we calculate the fletcher8 checksum to
match the online calculators (for example, [0]). This has two
important effects:

1) The maximum value it can return is now 0xEE rather than 0xFF.
   This is a documented limitation of the fletcher algorithm because
   it uses modulo operations and 0xF % 0xF is 0, not 0xF, so it's
   impossible to get a 0xF byte. Unfortunately, this does reduce the
   id space available.

2) Because the algorithm will not return 255, we can safely add one
   to the resulting value to ensure we don't end up with a VRID of
   0. 0 is not a valid VRID for keepalived, as reported in #21.

0: https://gchq.github.io/CyberChef/#recipe=Fletcher-8_Checksum()&input=bDN2dnFxbmktZjIxY2EtZG5z

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -13,8 +13,8 @@ import (
 func FletcherChecksum8(inp string) uint8 {
 	var ckA, ckB uint8
 	for i := 0; i < len(inp); i++ {
-		ckA = (ckA + inp[i]) & 0xf
-		ckB = (ckB + ckA) & 0xf
+		ckA = (ckA + inp[i]) % 0xf
+		ckB = (ckB + ckA) % 0xf
 	}
 	return (ckB << 4) | ckA
 }
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -159,9 +159,15 @@ func GetConfig(kubeconfigPath, clusterConfigPath string, apiVip net.IP, ingressV
 	// keepalived. This is safe because fletcher8 can never return 255 due to
 	// the modulo arithmetic that happens. The largest value it can return is
 	// 238 (0xEE).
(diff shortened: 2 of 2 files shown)


cybertron:
This is one possible approach. Another is to just clamp the id to > 0, but that way I think we'd double the chance of collisions at 1 because two hashes would be resolving to the same id. This approach slightly increases the chance of collisions across the board.

pkg/utils/utils.go
```diff
@@ -13,8 +13,8 @@ import (
 func FletcherChecksum8(inp string) uint8 {
 	var ckA, ckB uint8
 	for i := 0; i < len(inp); i++ {
-		ckA = (ckA + inp[i]) & 0xf
-		ckB = (ckB + ckA) & 0xf
+		ckA = (ckA + inp[i]) % 0xf
+		ckB = (ckB + ckA) % 0xf
```
cybertron: One other option would be to switch just this line back to &, which would allow us to get to a range of 0 to 254, which is exactly what we want. It wouldn't technically be fletcher 8 then, but I'm not sure how much we care about that.

mandre:
Thanks @cybertron for the patch, this is a nice way to take care of both issues reported in #21 and #22.

@celebdor any chance we can get this patch in? We're seeing the issue in CI occasionally.

mandre:
We should backport this to `release-4.2` branch.

We've seen it happen again in https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-openstack-serial-4.2/228

tomassedovic:
/backport release-4.2

(wonder if this works)

PR #25, opened 2019-10-01
Add EtcdBackends cluster member

cybertron:
This allows baremetal-runtimecfg to be used for generating the list
of etcd members for bootstrapping.


cybertron:
@celebdor I guess we still need you to approve this since you're the only approver in OWNERS.

PR #27, opened 2019-10-07
Monitor - add IPv6 support

yboaron:
Extend moniotr logic to support IPv6


celebdor:
@yboaron what is the best way to verify this?

yboaron:
@celebdor  I verified this change for both IPv4 and IPv6 in my env, for IPv6 rule, I did the following:
A. Changed the api-vip address in haproxy manifest see [1]
B. periodically force recreation of haproxy-monitor using [2] command  and verify that ip6table rule being created and deleted using [3]


[1]
  - name: haproxy-monitor
    securityContext:
      privileged: true
    image: quay.io/yboaron/bmruntime
    command:
    - monitor
    - "/etc/kubernetes/kubeconfig"
    - "/config/haproxy.cfg.tmpl"
    - "/etc/haproxy/haproxy.cfg"
    - "--api-vip"
    - "2001:0:9d3b:6ab8:1c48:3a1c:a95a:b1c2"

[2] sudo crictl rm    $(sudo crictl ps --name haproxy-monitor  | awk 'FNR==2{ print $1}') && sudo crictl rmi   quay.io/yboaron/bmruntime
[3] ]$ watch -n 5 'sudo ip6tables -t nat  --list  | grep 7443'

PR #28, opened 2019-10-22
Verify that Iptable rule being deleted

yboaron:
Due to multi-process implementation of the monitor code (it's a corner case),
 we may hit a race condition and as a result of this race, the iptable rule won't be deleted.
This fix verifies that the iptable rule is deleted.


celebdor:
@yboaron need to fix the go fmt

PR #29, opened 2019-11-10
Haproxy add support for IPv6 frontend

yboaron:
Add support for IPv6 frontend in Haproxy

### DIFF
--- a/test/data/haproxy.cfg.tmpl
+++ b/test/data/haproxy.cfg.tmpl
@@ -10,7 +10,7 @@ defaults
   timeout server       86400s
   timeout tunnel       86400s
 frontend  main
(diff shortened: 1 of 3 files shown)


test/data/haproxy.cfg.tmpl
```diff
@@ -10,7 +10,7 @@ defaults
   timeout server       86400s
   timeout tunnel       86400s
 frontend  main
-  bind :{{ .LBConfig.LbPort }}
+  bind {{ .LBConfig.FrontendAddr }}:{{ .LBConfig.LbPort }}
```
bcrochet: Does this notation work for an IPv6 address?

PR #31, opened 2019-11-14
Limit DNS upstream nameservers to 15

yboaron:
Verify that number of upstream servers passed to CoreDNS forward
plugin is maximum 15.


pkg/config/node.go
```diff
@@ -79,8 +79,8 @@ func getDNSUpstreams(resolvConfPath string) (upstreams []string, err error) {
 		case "nameserver":
 			// CoreDNS forward plugin takes up to 15 upstream servers
 			if len(fields) > 1 && len(upstreams) < 15 {
```
bcrochet: This would be a min of 2 and a max of 14...

PR #33, opened 2019-11-21
Bug 1775123: Prevent collisions between VRIDs

tomassedovic:
Because the fletcher8 algorithm only has a total of 255 hashes
available, it is highly likely that at some point we will have a
collision between our different VRIDs. This change adds checks for
that situation and increments the ids until they no longer collide.
This should be consistent across all nodes, so they will all end up
using the same ids for the same VIPs.


tomassedovic:
/close

Sorry, meant to target the 4.2 branch.

PR #34, opened 2019-11-21
Bug 1775123: Prevent collisions between VRIDs

tomassedovic:
Because the fletcher8 algorithm only has a total of 255 hashes
available, it is highly likely that at some point we will have a
collision between our different VRIDs. This change adds checks for
that situation and increments the ids until they no longer collide.
This should be consistent across all nodes, so they will all end up
using the same ids for the same VIPs.

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -155,9 +155,19 @@ func GetConfig(kubeconfigPath, clusterConfigPath string, apiVip net.IP, ingressV
 	node.Cluster.Name = clusterName
 	node.Cluster.Domain = clusterDomain
 
-	node.Cluster.APIVirtualRouterID = utils.FletcherChecksum8(node.Cluster.Name + "-api")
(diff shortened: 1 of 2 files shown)


tomassedovic:
@cybertron @yboaron please have a look. This is messing up the 4.2 CI for OpenStack.

Note that there were some git conflicts during cherry pick so I had to some parts manually.

mandre:
/lgtm

The 2 commits from https://github.com/openshift/baremetal-runtimecfg/pull/23/commits got squashed into one, but the change looks fine.

tomassedovic:
Oooh, that's why the cherry-pick failed! Sorry about that. I can forward them cleanly if y'all want.

cybertron:
For OpenStack I would generally have asked for them to be backported separately with cherry-pick -x to provide a breadcrumb trail for anyone looking through the git history in the future. I don't know what the policy in OpenShift is though.

The changes do look fine though and I agree this is good to backport.

mandre:
/close
This was merged in #35

PR #37, opened 2019-12-03
Add support for a provisioning VIP

bcrochet:
In some cases it may be necessary to specify a VIP that keepalived will then
listen on in order to provision within some networks. This adds support for
that interface and VIP.

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -22,17 +22,20 @@ import (
 var log = logrus.New()
 
 type Cluster struct {
-	Name                   string
-	Domain                 string
-	APIVIP                 string
-	APIVirtualRouterID     uint8
-	DNSVIP                 string
-	DNSVirtualRouterID     uint8
-	IngressVIP             string
-	IngressVirtualRouterID uint8
-	VIPNetmask             int
-	MasterAmount           int64
-	EtcdBackends           string
+	Name                           string
+	Domain                         string
+	APIVIP                         string
+	APIVirtualRouterID             uint8
+	DNSVIP                         string
+	DNSVirtualRouterID             uint8
+	IngressVIP                     string
+	IngressVirtualRouterID         uint8
+	APIProvisioningVIP             string
+	APIProvisioningVirtualRouterID uint8
+	VIPNetmask                     int
+	ProvVIPNetmask                 int
+	MasterAmount                   int64
+	EtcdBackends                   string
 }
 
 type Backend struct {
@@ -56,6 +59,7 @@ type Node struct {
 	ShortHostname     string
 	EtcdShortHostname string
 	VRRPInterface     string
+	VRRPProvInterface string
 	DNSUpstreams      []string
 }
 
@@ -144,7 +148,7 @@ func getClusterConfigMapInstallConfig(configPath string) (installConfig types.In
 	return ic, err
 }
 
-func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVip net.IP, ingressVip net.IP, dnsVip net.IP, apiPort, lbPort, statPort uint16) (node Node, err error) {
+func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVip net.IP, ingressVip net.IP, dnsVip net.IP, apiProvisioningVip net.IP, apiPort, lbPort, statPort uint16) (node Node, err error) {
 	// Try cluster-config.yml first
 	clusterName, clusterDomain, err := getClusterConfigClusterNameAndDomain(clusterConfigPath)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


bcrochet:
An MCO commit will be coming shortly to show how this will integrate.

test/data/keepalived.conf.tmpl
```diff
@@ -16,6 +16,12 @@ vrrp_script chk_ingress {
     weight 50
 }
 
+{{if .Cluster.APIProvisioningVIP}}
+vrrp_script chk_api_prov {
+    script "/usr/bin/curl -o /dev/null -kLs https://{{ .Cluster.APIProvisioningVIP }}:2379"
```
yboaron: Shouldn't the script check the local service (e.g: 127.0.0.1:2379)?

pkg/config/node.go
```diff
@@ -167,9 +171,13 @@ func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVip
 		node.Cluster.DNSVirtualRouterID++
 	}
 	node.Cluster.IngressVirtualRouterID = utils.FletcherChecksum8(node.Cluster.Name+"-ingress") + 1
-	for node.Cluster.IngressVirtualRouterID == node.Cluster.DNSVirtualRouterID || node.Cluster.IngressVirtualRouterID == node.Cluster.APIVirtualRouterID {
+	if node.Cluster.IngressVirtualRouterID == node.Cluster.DNSVirtualRouterID || node.Cluster.IngressVirtualRouterID == node.Cluster.APIVirtualRouterID {
 		node.Cluster.IngressVirtualRouterID++
 	}
+	node.Cluster.APIProvisioningVirtualRouterID = utils.FletcherChecksum8(node.Cluster.Name+"-api-provisioning") + 1
+	if node.Cluster.APIProvisioningVirtualRouterID == node.Cluster.IngressVirtualRouterID || node.Cluster.APIProvisioningVirtualRouterID == node.Cluster.DNSVirtualRouterID || node.Cluster.APIProvisioningVirtualRouterID == node.Cluster.APIVirtualRouterID {
```
yboaron: I don't think that it's mandatory to verify that PROV_VIP overlaps the rest of VIPs because it will run on another network interface.

bcrochet: The Router ID still needs to be unique, doesn't it?

test/data/keepalived.conf.tmpl
bcrochet: Yes. I'll change that.

pkg/config/node.go
yboaron: Ooop, I forgot to add Router ID to my previous comment. I think that Router IDs should be unique only for VRRP sessions that run on the same network interface, so since this VRRP session will run on a different network interface (the provisioning net interface) it isn't mandatory to verify that PROV router ID is unique

bcrochet: That's fair.

test/data/keepalived.conf.tmpl
```diff
@@ -16,6 +16,12 @@ vrrp_script chk_ingress {
     weight 50
 }
 
+{{if .Cluster.APIProvisioningVIP}}
+vrrp_script chk_api_prov {
+    script "curl -o /dev/null -kLs https://0:2379"
```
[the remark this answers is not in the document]
bcrochet: No. That was my bad. I've since changed it.

pkg/config/node.go
```diff
@@ -167,9 +171,10 @@ func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVip
 		node.Cluster.DNSVirtualRouterID++
 	}
 	node.Cluster.IngressVirtualRouterID = utils.FletcherChecksum8(node.Cluster.Name+"-ingress") + 1
-	for node.Cluster.IngressVirtualRouterID == node.Cluster.DNSVirtualRouterID || node.Cluster.IngressVirtualRouterID == node.Cluster.APIVirtualRouterID {
+	if node.Cluster.IngressVirtualRouterID == node.Cluster.DNSVirtualRouterID || node.Cluster.IngressVirtualRouterID == node.Cluster.APIVirtualRouterID {
```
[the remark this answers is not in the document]
bcrochet: Done.

pkg/config/node.go
```diff
@@ -204,6 +209,28 @@ func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVip
 		vips = append(vips, dnsVip)
 		node.Cluster.DNSVIP = dnsVip.String()
 	}
+	if apiProvisioningVip != nil {
+		node.Cluster.APIProvisioningVIP = apiProvisioningVip.String()
+	} else {
+		apiProvisioningVips, err := net.LookupIP("provisioning-api." + node.Cluster.Name + node.Cluster.Domain)
```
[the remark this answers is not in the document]
bcrochet: We were not planning to add it to openshift/api.

PR #38, opened 2019-12-20
Convert /128 addresses to /64

derekhiggins:
In order to select an interface for the VIPs we
are trying to find a interface with an IP in the
same subnet.

If using DHCPv6 then addresses have a /128 address,
assume in this case its enough to test if the VIP
is in the same /64.

### DIFF
--- a/pkg/config/net.go
+++ b/pkg/config/net.go
@@ -3,6 +3,7 @@ package config
 import (
 	"fmt"
 	"net"
+	"strings"
 )
 
 func getInterfaceAndNonVIPAddr(vips []net.IP) (vipIface net.Interface, nonVipAddr *net.IPNet, err error) {
@@ -27,6 +28,11 @@ func getInterfaceAndNonVIPAddr(vips []net.IP) (vipIface net.Interface, nonVipAdd
 				if _, ok := vipMap[n.IP.String()]; ok {
 					continue // This is a VIP, let's skip
 				}
+				// FIXME: Breaks if more then one interface on the host has the same prefix
+				// DHCPv6 assigns addresses with a /128 if encountered assume /64
+				// so that the n.Contains returns true if the VIP has the same prefix
+				_, n, _ = net.ParseCIDR(strings.Replace(addr.String(), "/128", "/64", 1))
+
 				if n.Contains(vips[0]) {
 					return iface, n, err
 				}


pkg/config/net.go
```diff
@@ -27,6 +27,11 @@ func getInterfaceAndNonVIPAddr(vips []net.IP) (vipIface net.Interface, nonVipAdd
 				if _, ok := vipMap[n.IP.String()]; ok {
 					continue // This is a VIP, let's skip
 				}
+				// FIXME: Breaks if more then one interface on the host has the same prefix
```
cybertron: Hmm, is there any way we could look this up? On the bootstrap I think we could get it from the cluster config since that has the machinecidr value, but I'm not sure about the masters. Maybe MCO could dump it out for us?

hardys:
@derekhiggins currently this fails to build because we're missing the import of the strings module, can you add that pls so we can use this as a workaround for ipv6 testing?

@cybertron @celebdor we need to decide the way forward for this as it's going to block testing of ipv6 controlplane, care to revisit this review so we can figure out the preferred fix for this?

derekhiggins:
> @derekhiggins currently this fails to build because we're missing the import of the strings module, can you add that pls so we can use this as a workaround for ipv6 testing?

oops, done

cybertron:
I'm okay with merging this as-is to unblock ipv6 testing, but I assume we can't hard-code /64 indefinitely, right? Or is the ipv6 address space so large that everyone just uses /64?

hardys:
> I'm okay with merging this as-is to unblock ipv6 testing, but I assume we can't hard-code /64 indefinitely, right? Or is the ipv6 address space so large that everyone just uses /64?

I think the issue is the subnet/prefix isn't provided by default for dhcpv6, so clients have to configure a 128 in the absence of that prefix information.  I found https://tools.ietf.org/html/rfc5942 which has some more details about it.

It seems there are ways to influence the DHCP server to override this behavior and advertise the prefix/subnet but I've not yet figured out how to do that with dnsmasq.

This PR seems like a reasonable workaround in the meantime but clearly it's not the most robust solution so we should look for a better alternative.

russellb:
@cybertron can you /approve ?

celebdor:
/lgtm

We'll need to find a better way soon. Let's unblock this for now.

russellb:
@celebdor can you help get the 4.3 backport merged as well?  Thanks!

PR #41, opened 2020-01-16
Revert "Convert /128 addresses to /64"

derekhiggins:
This reverts commit 24845d3960ce1dcadac2dfaaac089783ba2aace7.

Passing an ipaddress through net.ParseCIDR ends up stripping
out averything but the network part of the IPaddress.

### DIFF
--- a/pkg/config/net.go
+++ b/pkg/config/net.go
@@ -3,7 +3,6 @@ package config
 import (
 	"fmt"
 	"net"
-	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


mandre:
Yes, please! https://github.com/openshift/baremetal-runtimecfg/commit/24845d3960ce1dcadac2dfaaac089783ba2aace7 broke our openstack jobs and we were looking into it as well.

cybertron:
/lgtm
/approve

I posted some code that should fix this in the issue, but in the meantime let's do this to unblock everyone.

PR #44, opened 2020-02-09
Bug 1790823: Consider mdns hostname file existence in utils.ShortHostname

yboaron:
The utils.ShortHostname function should read hostname from mdns
hostname, if mdns hostname file exists.

This mdns hostname file creation is handled by [1] MCO PR.

[1] https://github.com/openshift/machine-config-operator/pull/1446crea

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -8,7 +8,11 @@ import (
 	"net/http"
 	"os"
 	"strings"
+
+	"github.com/sirupsen/logrus"
 )
+
+var log = logrus.New()
 
 func FletcherChecksum8(inp string) uint8 {
 	var ckA, ckB uint8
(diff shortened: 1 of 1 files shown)


celebdor:
@yboaron needs to be tied to the same bug as https://github.com/openshift/machine-config-operator/pull/1446 (both in bugzilla and here in the title of the PR

pkg/utils/utils.go
```diff
@@ -20,10 +26,27 @@ func FletcherChecksum8(inp string) uint8 {
...
-	if err != nil {
-		panic(err)
+	var hostname string
+
+	if _, err := os.Stat(mdns_hostname_file); err == nil {
+		dat, err := ioutil.ReadFile(mdns_hostname_file)
+		if err != nil {
+			panic(err)
+		}
+		hostname = string(dat[:len(dat)-1])
```
celebdor: What is the character that you are trimming? `\n` ? If so, it is nicer to use https://golang.org/pkg/strings/#TrimSpace

pkg/utils/utils.go
```diff
@@ -20,9 +24,29 @@ func FletcherChecksum8(inp string) uint8 {
...
-		panic(err)
+	var hostname string
+
+	if filePath, ok := os.LookupEnv("RUNTIMECFG_HOSTNAME_PATH"); ok {
+		dat, err := ioutil.ReadFile(filePath)
+		if err != nil {
+			log.WithFields(logrus.Fields{
+				"filePath": filePath,
+			}).Error("Failed to read file")
+			panic(err)
```
celebdor: Why do you panic? The signature of this method states that it can return an error. Thus, the expected statement here would be 
```suggestion
			return "", err
```

pkg/utils/utils.go
```diff
@@ -20,9 +24,29 @@ func FletcherChecksum8(inp string) uint8 {
...
+		}
+		hostname = strings.TrimSuffix(string(dat[:]), "\n")
```
celebdor: why is the `[:]` necessary?
Why not just run TrimSpace?

pkg/utils/utils.go
```diff
@@ -20,9 +24,29 @@ func FletcherChecksum8(inp string) uint8 {
...
+		log.WithFields(logrus.Fields{
+			"hostname": hostname,
+			"file":     filePath,
+		}).Info("Hostname retrieved from a file")
```
celebdor: Maybe that's more for a debug log

pkg/utils/utils.go
yboaron: My bad, thanks!

PR #46, opened 2020-02-14
Add non-virtual-ip script

vrutkovs:
MCO is placing a script on every node to setup non-virtual-ip via NM (see https://github.com/openshift/machine-config-operator/blob/fcos/templates/worker/00-worker/openstack/files/openstack-non-virtual-ip.yaml)

This script runs on host and requires Python on host. Instead this script should be in `baremetalcfg-runtime` container. MCO should use a podman wrapper with `--net=host` to run it in container, as some hosts (e.g. Fedora CoreOS) may not have python installed.

Eventually this script should be rewritten in golang.

Reference: https://github.com/openshift/machine-config-operator/issues/1430

### DIFF
--- a/scripts/nodeip-finder
+++ b/scripts/nodeip-finder
@@ -1,0 +1,73 @@
+#!/usr/libexec/platform-python
+# /* vim: set filetype=python : */
+"""Writes Kubelet and CRI-O configuration to choose the right IP address
+
+For kubelet, a systemd environment file with a KUBELET_NODE_IP setting
+For CRI-O it drops a config file in /etc/crio/crio.conf.d"""
+from importlib import util as iutil
+from importlib import machinery as imachinery
+from types import ModuleType
+import os
+import pathlib
+import socket
+import sys
+
+loader = imachinery.SourceFileLoader(
+    'non_virtual_ip',
+    os.path.join(os.path.dirname(os.path.realpath(__file__)), 'non_virtual_ip'))
+spec = iutil.spec_from_loader('non_virtual_ip', loader)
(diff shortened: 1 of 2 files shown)


metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1516/

vrutkovs:
This script is not yet being used anywhere, so its safe to merge

vrutkovs:
Tested this on OKD - workers join the cluster as expected.

/cc @celebdor 

PTAL.

mandre:
This script was recently modified by commits https://github.com/openshift/machine-config-operator/commit/cdd53fcb01cf6a091b2e3fe99f8c101a160fd558 and https://github.com/openshift/machine-config-operator/commit/d93d1e97d3c4914633a41c1139cac89a1c0e507c.

As of yesterday with https://github.com/openshift/machine-config-operator/pull/1483, OpenStack also uses this newer `non-virtual-ip` script. The same PR introduced another python script (sorry) that should be added to baremetal-runtimecfg the same way.

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1526/

mandre:
/lgtm

The scripts now functionally match the BM and OpenStack ones in master MCO. There are some whitespace diffs that will make pep8 complain, but I do not think we care here, do we?

metal3ci:
Build FAILURE, see build http://10.8.144.11:8080/job/dev-tools/1527/

vrutkovs:
/retest

Prow issues

PR #47, opened 2020-02-20
Add unicast IP server

bcrochet:
Add a simple server that will return the IP address of the bootstrap server.
One should contact it at the API VIP on port 64444.


pkg/config/node.go
```diff
@@ -144,6 +152,41 @@ func getClusterConfigMapInstallConfig(configPath string) (installConfig types.In
 	return ic, err
 }
 
+func getBootstrapIP(apiVip string) (bootstrapIP string, err error) {
+	conn, err := net.Dial("tcp", net.JoinHostPort(apiVip, "64444"))
```
yboaron: Do we need o handle the add-master use case?
And I think it will safer to use here DialTimeout instead of Dial

pkg/monitor/dynkeepalived.go
```diff
@@ -42,7 +42,7 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
 			if err != nil {
 				return err
 			}
-			if prevConfig == nil || prevConfig.VRRPInterface != newConfig.VRRPInterface {
+			if prevConfig == nil || prevConfig.VRRPInterface != newConfig.VRRPInterface || len(prevConfig.LBConfig.Backends) != len(newConfig.LBConfig.Backends) || prevConfig.BootstrapIP != newConfig.BootstrapIP {
```
yboaron: Do we want to update keepalived.conf after a single poll or we would like to add stabilization here?
What will happen If the DNS ( we retrieve backends from DNS server) won't reply for X seconds?

pkg/utils/utils.go
```diff
@@ -24,11 +24,16 @@ func ShortHostname() (shortName string, err error) {
 	if err != nil {
 		panic(err)
 	}
-	splitHostname := strings.SplitN(hostname, ".", 2)
-	shortName = splitHostname[0]
+	shortName, err = GetShortHostname(hostname)
 	return shortName, err
 }
 
+func GetShortHostname(hostName string) (shortName string, err error) {
```
yboaron: Does return error is needed here? function always returns error=nil

pkg/config/node.go
bcrochet: getBootstrapIP will just return empty string. net.Dial is blocking. If you mean that a new master comes online, as long as the master has a SVC entry, it should be picked up.

pkg/monitor/dynkeepalived.go
bcrochet: We could put in a counter that if that section has changed for 3 polls, then go ahead and commit the change.

bcrochet: We could put in a counter that if that section has changed for 3 polls, then go ahead and commit the change.

yboaron: Sounds good, we did something similar in haproxy/monitor

pkg/config/node.go
bcrochet: My bad. I meant that yes, it blocks, but not if there is a connection refused, which is what will happen once the bootstrap goes down. Then it just returns immediately. DialTimeout probably should be used anyway.

pkg/config/node.go
```diff
@@ -222,6 +303,16 @@ func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVip
...
+	node.BootstrapIP, err = getBootstrapIP(apiVip.String())
+	if err != nil {
+		log.Warn(err)
+	}
+
+	node.IngressConfig, err = GetIngressConfig(kubeconfigPath)
+	if err != nil {
+		log.Warn(err)
+	}
+
```
yboaron: 1. getBootstrapIP & GetIngressConfig functions required only for keepalived-monitor, maybe we should consider split them into separate function or add a boolean flag to Getconfig function.
2. Another point, there's no meaning calling to 'getBootstrapIP' every poll interval forever after bootstrap was deleted. 
One side effect of that is keepalived-moniotor log is full with:
"time="2020-03-03T11:11:56Z" level=info msg="An error occurred on dial: dial tcp 192.168.111.5:64444: connect: connection refused"
time="2020-03-03T11:11:56Z" level=warning msg="dial tcp 192.168.111.5:64444: connect: connection refused" "

pkg/monitor/dynkeepalived.go
```diff
@@ -42,7 +44,13 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
...
 			}
-			if prevConfig == nil || prevConfig.VRRPInterface != newConfig.VRRPInterface {
+			curConfig = &newConfig
+			if appliedConfig == nil || !cmp.Equal(*appliedConfig, *curConfig) {
+				if prevConfig == nil || cmp.Equal(*prevConfig, *curConfig) {
+					configChangeCtr++
+				} else {
+					configChangeCtr = 1
+				}
 				log.WithFields(logrus.Fields{
```
yboaron: Shouldn't you need to apply changes only if configChangeCtr > some_threshold?

pkg/monitor/dynkeepalived.go
```diff
@@ -42,7 +44,13 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
...
 				return err
 			}
-			if prevConfig == nil || prevConfig.VRRPInterface != newConfig.VRRPInterface {
+			curConfig = &newConfig
+			if appliedConfig == nil || !cmp.Equal(*appliedConfig, *curConfig) {
+				if prevConfig == nil || cmp.Equal(*prevConfig, *curConfig) {
+					configChangeCtr++
+				} else {
+					configChangeCtr = 1
+				}
```
yboaron: Shouldn't you need to apply changes only if configChangeCtr > some_threshold?

pkg/config/node.go
bcrochet: 1. Fixed.
2. I'll try to figure out a way to reliably detect that the bootstrap is no more. For now, I'd like to get this landed.

pkg/monitor/dynkeepalived.go
bcrochet: We can address the second part in another pull.

pkg/utils/utils.go
bcrochet: I removed that.

pkg/config/node.go
```diff
@@ -36,9 +39,11 @@ type Cluster struct {
 }
 
 type Backend struct {
-	Host    string
-	Address string
-	Port    uint16
+	Host               string
+	CanonicalHost      string
+	CanonicalShortName string
```
yboaron: Do we use CanonicalHost and CanonicalShortName ?

bcrochet: Yes. It's being used in the templates.

yboaron: I "greped" MCO and couldn't find these two fields. what am I missing here?

celebdor:
We're keeping vrrp for keepalived as the default for now. Let's shelve this for now
/hold

PR #51, opened 2020-03-11
Bug 1812526: [release-4.4]  Add support for coredns watcher

yboaron:
Since  Corefile rendering for the Coredns container runs as initcontainer, changes to /etc/resolv.conf that happened after Coredns rendering completed won't be applied to Coredns Corefile.
With coredns-monitor we should be able to watch those changes and update Corefile correspondingly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ network_closure.sh
 runtimecfg
 monitor
 dynkeepalived
+corednsmonitor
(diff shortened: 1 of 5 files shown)


yboaron:
I run 4.4 deployment with coredns-monitor PRs - looks OK

ashcrow:
This is likely blocking https://github.com/openshift/machine-config-operator/pull/1557

celebdor:
This should link to the same bugzilla in title as the PR that depends on it.

yboaron:
/retitle Bug 1812526: [release-4.4]  Add support for coredns watcher

PR #52, opened 2020-03-22
Retrieve HAProxy members using KUBE-API

yboaron:
Update HAProxy monitor to retrieve LB backends details using
KUBE-API instead of  _etcd-server-ssl._tcp SRV record.


yboaron:
This PR should be tested together with https://github.com/openshift/machine-config-operator/pull/1574 MCO PR

pkg/config/node.go
```diff
@@ -226,52 +226,55 @@ func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVip
...
-			continue
+	for _, node := range nodes.Items {
+		masterIp = ""
+		for _, address := range node.Status.Addresses {
+			if address.Type == v1.NodeInternalIP {
+				masterIp = address.Address
+				break
+			}
+		}
+		if masterIp != "" {
```
cybertron: Should we log if this check fails? Seems like something would be wrong if that's the case.

pkg/config/node.go
```diff
@@ -226,52 +226,55 @@ func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVip
...
 	}
-
-	for _, srv := range srvs {
-		addr, err := utils.GetFirstAddr(srv.Target)
-		if err != nil {
-			log.WithFields(logrus.Fields{
-				"member": srv.Target,
-			}).Error("Failed to get address for member")
-			continue
+	for _, node := range nodes.Items {
```
celebdor: masterIp is not used until this block here, let's move its declaration closer to the using code. Like just before 258

pkg/config/node.go
celebdor: Agreed

pkg/config/node.go
```diff
@@ -32,7 +33,6 @@ type Cluster struct {
 	IngressVirtualRouterID uint8
 	VIPNetmask             int
 	MasterAmount           int64
-	EtcdBackends           string
```
celebdor: Is nothing else using this field anymore? It seems to me like having the masters listed in the cluster struct is nice (though maybe the field name is not fitting).

pkg/config/node.go
yboaron: Done

pkg/config/node.go
yboaron: Done

pkg/config/node.go
yboaron: AFAIK no one is using this field anymore. I'll double-check that.
We have the masters list in LB.backends, do you think we should hold it also in cluster struct?

pkg/monitor/monitor.go
```diff
@@ -27,7 +26,7 @@ type RuntimeConfig struct {
 	LBConfig *config.ApiLBConfig
 }
 
-func Monitor(clusterName, clusterDomain, templatePath, cfgPath, apiVip string, apiPort, lbPort, statPort uint16, interval time.Duration) error {
+func Monitor(kubeconfigPath, clusterName, clusterDomain, templatePath, cfgPath, apiVip string, apiPort, lbPort, statPort uint16, interval time.Duration) error {
```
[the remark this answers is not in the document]
yboaron: Yep, it's an API change, but we should be safe with this change since we have only internal consumers (well it's a single consumer: cmd/monitor) and we updated both consumer and provider.

pkg/monitor/monitor.go
```diff
@@ -50,17 +49,20 @@ func Monitor(clusterName, clusterDomain, templatePath, cfgPath, apiVip string, a
...
 			return nil
 		default:
-			config, err := config.GetLBConfig(domain, apiPort, lbPort, statPort, net.ParseIP(apiVip))
+			config, err := config.GetLBConfig(kubeconfigPath, apiPort, lbPort, statPort, net.ParseIP(apiVip))
 			if err != nil {
-				return err
+				log.WithFields(logrus.Fields{
+					"kubeconfigPath": kubeconfigPath,
+				}).Info("GetLBConfig failed, sleep half of interval and retry")
+				time.Sleep(interval / 2)
```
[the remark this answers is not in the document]
yboaron: I noticed that haproxy-moniotr container exits (and restarted by kubelet) , the root cause was  'GetLBConfig' failed to read nodes via OCP API ( error was something like 'etcd leader changed').
So, since we implemented a de-bounce threshold in the monitor code, I think we should ignore this poll instead of exiting, and also it make sense to start the next poll faster. I set it to half the default interval, but I'm totally open for other ideas.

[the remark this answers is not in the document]
cybertron: Yeah, I understand why a retry is needed here, I just wasn't sure why we needed a shorter interval. I'm not concerned about it, but I wanted to see if I was missing something.

yboaron:
I can see that tests fail also for other PRs, so I guess it isn't my fault  :-)

PR #53, opened 2020-03-26
scripts: Add python2 variants of python scripts

LorbusChris:
https://pypi.org/project/3to2/ was used for translation.

This is so the scripts can be run within a UBI7 container.
Retains the Python3 scripts for future use within UBI8 containers.

Translation Warnings:
- nodeip-finder-py2:
```
RefactoringTool: Warnings/messages while refactoring:
RefactoringTool: ### In file nodeip-finder-py2 ###
RefactoringTool: Line 21: Removing function annotations completely.
```

- non-virtual-ip-py2:
```
RefactoringTool: Warnings/messages while refactoring:
RefactoringTool: ### In file non-virtual-ip-py2 ###
RefactoringTool: Line 20: Removing function annotations completely.
```

/cc @vrutkovs

### DIFF
--- a/scripts/nodeip-finder-py2
+++ b/scripts/nodeip-finder-py2
@@ -1,0 +1,62 @@
+#!/usr/libexec/platform-python
+# /* vim: set filetype=python : */
+u"""Writes Kubelet and CRI-O configuration to choose the right IP address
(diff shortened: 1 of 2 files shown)


LorbusChris:
this isn't needed after all (we'll use go implementation in MCD for OKD).
/close

PR #54, opened 2020-04-07
bug 1821667: runtimecfg: tool to show the Virtual Router IDs

celebdor:
With this tool, an operator will be able to easily determine before
deployment if the cluster names chosen will end up with conflicting
Virtual Router IDs. One can't use the same Virtual Router ID for
different VIPs in the same multicast domain.

Usage example:

    $ ./runtimecfg vr-ids cnf10
    APIVirtualRouterID: 147
    DNSVirtualRouterID: 158
    IngressVirtualRouterID: 2
    $ ./runtimecfg vr-ids cnf11
    APIVirtualRouterID: 228
    DNSVirtualRouterID: 239
    IngressVirtualRouterID: 147

In this example, we can see that cluster cnf10 would end up with the
API virtual router id conflicting with cnf11's ingress router id. In
this case, if both clusters share the same control plane multicast
domain, different cluster names should be chosen.

### DIFF
--- a/cmd/runtimecfg/runtimecfg.go
+++ b/cmd/runtimecfg/runtimecfg.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"reflect"
+	"strings"
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


yboaron:
code looks good, I'll LGTMEd after I verify that deployment with baremetal image

cybertron:
/test e2e-metal-ipi

Note that we have a ci job on this repo now so manual verification shouldn't be needed.

The change also lgtm.

yboaron:
run deployment locally - looks OK.

yboaron:
/retitle bug 1821667: runtimecfg: tool to show the Virtual Router IDs

PR #56, opened 2020-04-14
DNM: Test ci on bad patch

cybertron:
Just verifying that ci fails correctly on a broken change.


cybertron:
Looks like this failed correctly. Since the api doesn't come up, there are no container logs to tell us what actually failed but this is what I would expect from the error I introduced.

PR #59, opened 2020-04-28
Bug 1821720: Move haproxy listening port due to conflict

mandre:
The "Recovering from expired control plane certificates" procedure [1]
stands up a recovery API server on port 7443 [2], conflicting with the
port the haproxy process is listening on.

This patch moves the haproxy port to 9443 instead.

[1] https://docs.openshift.com/container-platform/4.3/backup_and_restore/disaster_recovery/scenario-3-expired-certs.html

[2] https://github.com/openshift/cluster-kube-apiserver-operator/blob/3161546a248f20eb67231017d0ae43c245bfa4cb/pkg/recovery/apiserver.go#L307

### DIFF
--- a/cmd/monitor/monitor.go
+++ b/cmd/monitor/monitor.go
@@ -52,7 +52,7 @@ func main() {
 		},
 	}
 	rootCmd.Flags().Uint16("api-port", 6443, "Port where the OpenShift API listens at")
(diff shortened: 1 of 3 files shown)


cybertron:
/lgtm
/retest

Job failed deploying masters, which seems unlikely to be related to this.

PR #60, opened 2020-04-29
Add A/AAAA record types for VIP DNS entries

cybertron:
We add records for some of the VIPs to our internal coredns server.
These records need to be of the appropriate type for the ip version
used. This change adds record type fields to the cluster for use
in the coredns config file. Two fields are added for each VIP:
One that specifies the correct type of record, and one that specifies
the other so that we are able to correctly handle requests for the
wrong record type too.

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -28,10 +28,14 @@ type Cluster struct {
 	Domain                 string
 	APIVIP                 string
 	APIVirtualRouterID     uint8
+	APIVIPRecordType       string
+	APIVIPEmptyType        string
 	DNSVIP                 string
 	DNSVirtualRouterID     uint8
 	IngressVIP             string
 	IngressVirtualRouterID uint8
+	IngressVIPRecordType   string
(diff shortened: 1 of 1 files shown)


pkg/config/node.go
```diff
@@ -202,13 +202,25 @@ func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVip
 	}
 
 	vips := make([]net.IP, 0)
+	node.Cluster.APIVIPRecordType = "A"
+	node.Cluster.APIVIPEmptyType = "AAAA"
```
celebdor, line 206: Shouldn't these fields be added to the struct in line 26?

cybertron: Somebody forgot to commit his local changes...

cybertron:
/retest

packet host never came up

cybertron:
/retest

It's failing progressively faster each time...

cybertron:
/retest

Never deployed masters

PR #62, opened 2020-05-11
Bug 1823950: Switch to /readyz for haproxy healthchecking

cybertron:
Per [0], the /readyz endpoint is how the api communicates that it
is gracefully shutting down. Once /readyz starts to report failure,
we want to stop sending traffic to that backend. If we wait for
/healthz, it may be too late because once /healthz starts failing
the api is already not accepting connections.

0: https://github.com/openshift/installer/pull/3537


cybertron:
/hold

I'm not positive we want to change this here. The priority for fixing the bug is the MCO patch: https://github.com/openshift/machine-config-operator/pull/1724

cybertron:
Included in #68.

PR #63, opened 2020-05-26
Bug 1840222: Fail if no upstream DNS servers are found

cybertron:
Coredns will fail to start if we configure the forward plugin with
no upstream servers. The init container responsible for rendering
the config needs to fail if that situation arises so it can be
retried later when hopefully resolv.conf has been populated correctly.

We're not sure how it happens that there are no upstream servers,
but it's clearly wrong for us to render an invalid config so this
is the safe thing to do anyway.

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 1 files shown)


stbenjam:
CentOS mirrors...
```
[MIRROR] git-core-doc-2.18.2-2.el8_1.noarch.rpm: Curl error (7): Couldn't connect to server for http://mirror.tocici.com/centos/8.1.1911/AppStream/x86_64/os/Packages/git-core-doc-2.18.2-2.el8_1.noarch.rpm [Failed to connect to mirror.tocici.com port 80: Connection timed out] 
```

/test e2e-metal-ipi

stbenjam:
/hold

Please don't lgtm yet, want to see how e2e-metal-ipi job does

stbenjam:
Passed! That's encouraging.

/hold cancel
/lgtm

PR #67, opened 2020-06-09
Update LB backends retrieval code to failover to localhost:kube-apiserver

yboaron:
This PR updates LB backends details retrieval code to failover to localhost:kube-apiserver.
Failover to localhost:kube-apiserver also helps to avoid the race between haproxy-monitor and NM resolv prepend script, for more information check [1].

 [1] https://bugzilla.redhat.com/show_bug.cgi?id=1849432

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -22,6 +22,8 @@ import (
 	"github.com/openshift/installer/pkg/types"
 )
 
+const localhostKubeApiServerUrl string = "https://localhost:6443"
+
 var log = logrus.New()
 
 type Cluster struct {
@@ -261,9 +263,15 @@ func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVip 
 	return node, err
 }
 
+// getSortedBackends builds config to communicate with kube-api based on kubeconfigPath parameter value, if kubeconfigPath is not empty it will build the
+// config based on that content else config will point to localhost.
 func getSortedBackends(kubeconfigPath string) (backends []Backend, err error) {
 
-	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+	kubeApiServerUrl := ""
+	if kubeconfigPath == "" {
+		kubeApiServerUrl = localhostKubeApiServerUrl
+	}
+	config, err := clientcmd.BuildConfigFromFlags(kubeApiServerUrl, kubeconfigPath)
 	if err != nil {
 		log.WithFields(logrus.Fields{
 			"err": err,
@@ -313,17 +321,23 @@ func GetLBConfig(kubeconfigPath string, apiPort, lbPort, statPort uint16, apiVip
 		LbPort:   lbPort,
 		StatPort: statPort,
 	}
+
 	// LB frontend address: IPv6 '::' , IPv4 ''
 	if apiVip.To4() == nil {
 		config.FrontendAddr = "::"
 	}
-
+	// Try reading master nodes details first from api-vip:kube-apiserver and failover to localhost:kube-apiserver
 	backends, err := getSortedBackends(kubeconfigPath)
 	if err != nil {
-		log.WithFields(logrus.Fields{
-			"kubeconfigPath": kubeconfigPath,
-		}).Error("Failed to retrieve API members information")
-		return config, err
+		log.Infof("An error occurred while trying to read master nodes details from api-vip:kube-apiserver: %v", err)
+		log.Infof("Trying to read master nodes details from localhost:kube-apiserver")
+		backends, err = getSortedBackends("")
+		if err != nil {
+			log.WithFields(logrus.Fields{
+				"kubeconfigPath": kubeconfigPath,
+			}).Error("Failed to retrieve API members information")
+			return config, err
+		}
 	}
 
 	// The backends port is the Etcd one, but we need to loadbalance the API one


pkg/config/node.go
```diff
@@ -307,18 +309,34 @@ func getSortedBackends(kubeconfigPath string) (backends []Backend, err error) {
 	return backends, err
 }
 
-func GetLBConfig(kubeconfigPath string, apiPort, lbPort, statPort uint16, apiVip net.IP) (ApiLBConfig, error) {
+func GetLBConfig(kubeconfigPath string, apiPort, lbPort, statPort uint16, apiVip net.IP, queyrAlsoApiVipOpt ...bool) (ApiLBConfig, error) {
```
cybertron, line 312: Typo in the variable name: s/queyr/query/

cybertron: Also, can we skip the variadic workaround for Go's limitations and just always pass this? In fact, does it even need to be configurable? Are there circumstances where we need to _not_ query the api vip?

pkg/config/node.go
```diff
@@ -307,18 +309,34 @@ func getSortedBackends(kubeconfigPath string) (backends []Backend, err error) {
...
+	// api-vip:kube-apiserver.
+	backends, err := getSortedBackends(kubeconfigPath, localhostKubeApiServerUrl)
+	if err != nil {
+		log.Infof("An error occurred while trying to read master nodes details from localhost:kube-apiserver: %v", err)
+		if queyrAlsoApiVip {
+			log.Infof("Trying to read master nodes details from api-vip:kube-apiserver")
+			backends, err = getSortedBackends(kubeconfigPath, "")
+		}
+	}
 	if err != nil {
```
cybertron, line 340: I find this logic a bit unintuitive because of the way it splits the err check from the call that produces it. Can we just move this into the block with the second getSortedBackends call?

yboaron:
/retitle Bug 1849432: allow haproxy-monitor to retrieve nodes details when resolv.conf is not synced with host's resolv.conf.

pkg/config/node.go
```diff
@@ -261,9 +263,9 @@ func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVip
 	return node, err
 }
 
-func getSortedBackends(kubeconfigPath string) (backends []Backend, err error) {
+func getSortedBackends(kubeconfigPath string, kubeApiServerUrl string) (backends []Backend, err error) {
```
bcrochet, line 266: Nit: If there isn't presently any other code that would break due to a change in the order here, I'd rather see the kubeApiServerUrl param come first, to match the order of BuildConfigFromFlags. It's a small nit, but I had to say it. :)

pkg/config/node.go
```diff
@@ -313,17 +315,30 @@ func GetLBConfig(kubeconfigPath string, apiPort, lbPort, statPort uint16, apiVip
...
+	// and only failover to api-vip:kube-apiserver
+	backends, err := getSortedBackends(kubeconfigPath, localhostKubeApiServerUrl)
 	if err != nil {
-		log.WithFields(logrus.Fields{
-			"kubeconfigPath": kubeconfigPath,
-		}).Error("Failed to retrieve API members information")
-		return config, err
+		log.Infof("An error occurred while trying to read master nodes details from localhost:kube-apiserver: %v", err)
+		log.Infof("Trying to read master nodes details from api-vip:kube-apiserver")
+		backends, err = getSortedBackends(kubeconfigPath, "")
```
bcrochet, line 335: This seems rather magical. Why does passing "" mean api-vip:kube-apiserver? The casual reader would not know how this works. Is there a more explicit value we could pass?

yboaron: I added a Docstring to getSortedBackends function, hope that now it's more explicit.

pkg/config/node.go
yboaron: Found another solution, I deleted the 'kubeApiServerUrl' parameter. :-)

pkg/config/node.go
```diff
@@ -313,17 +321,30 @@ func GetLBConfig(kubeconfigPath string, apiPort, lbPort, statPort uint16, apiVip
...
-	backends, err := getSortedBackends(kubeconfigPath)
+	// since:
+	// 1: Keepalived script for Api vip checks HAProxy LB status in addition to localhost-kube-apiserver.
+	// 2: Masters nodes details needed for HAProxy config retrieved from kube-apiserver .
+	// to break the circular dependency we should first read masters details from localhost:kube-apiserver instead of
+	// api-vip:kube-apiserver.
+	// Since master nodes details are used (in unicast mode) in api VRRP instance at keeplived config
+	// we should first retrieve the information from localhost:kube-apiserver (to avoid circular dependency),
+	// and only failover to api-vip:kube-apiserver
+	backends, err := getSortedBackends("")
```
cybertron, line 337: This feels backward to me. Other than the initial deployment where the VIP isn't up yet, the normal case is going to be that the API VIP is responding and we want to use that. If we check localhost first, then every time the local apiserver goes down for any reason we'll spam the logs with these messages. I think I'd prefer to do the VIP check first and fall back to localhost only if it isn't available. What do you think?

pkg/config/node.go
```diff
@@ -261,9 +263,15 @@ func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVip
 	return node, err
 }
 
+// getSortedBackends build config to communicate with kube-api based on kubeconfigPath parameter value, if kubeconfigPath is not empty it will build the
```
cybertron, line 266: Ick, I dislike overloading variable meanings this way, but I guess client-go already did the same thing so we might as well go with it. :-/

pkg/config/node.go
```diff
@@ -313,17 +321,30 @@ func GetLBConfig(kubeconfigPath string, apiPort, lbPort, statPort uint16, apiVip
...
 	}
+
 	// LB frontend address: IPv6 '::' , IPv4 ''
 	if apiVip.To4() == nil {
 		config.FrontendAddr = "::"
 	}
-
-	backends, err := getSortedBackends(kubeconfigPath)
+	// since:
+	// 1: Keepalived script for Api vip checks HAProxy LB status in addition to localhost-kube-apiserver.
```
cybertron, line 330: I don't know that these comments are still relevant. The loadbalanced healthchecks work fine with the API VIP. Also, the current version of them is smart enough to handle the case where no haproxy instances are running but a local apiserver is.

pkg/config/node.go
yboaron: Sounds like a good idea if we only need to handle HAProxy backends details retrieval, but since this code will be used also for retrieving keepalived API_VIP group members (unicast mode), I'm not sure it's a good idea to have that circular dependency.
I can change the order per your recommendation for this PR and will update this in the unicast PR.

yboaron:
/retitle Update LB backends retrieval code to failover to localhost:kube-apiserver

yboaron:
This commit was merged with [https://github.com/openshift/baremetal-runtimecfg/pull/65](https://github.com/openshift/baremetal-runtimecfg/pull/65) 

/close

PR #68, opened 2020-06-11
Bug 1847086 : Update kube-api health check to use localhost

yboaron:
K8s health check should use localhost instead of 127.0.0.1 to support both
IPv6 and IPv4


yboaron:
/retitle Bug 1847086 : Update kube-api health check to use localhost

cybertron:
This should be good. I proposed the readyz change in #62, but it wasn't clear if we needed it. I don't think it will hurt to switch though.

metal ci is broken right now but once it's working again we can retest and verify that this passes.

PR #70, opened 2020-07-02
Add check for iptables rule to keepalived-monitor

cybertron:
In order to have keepalived use the loadbalanced api endpoint, we need to know whether the iptables rule to redirect traffic to haproxy is present. Since the keepalived container doesn't have the necessary bits to work with iptables itself, we can instead do it in the monitor container and just use a file to indicate whether the rule is present. This also allows us to reuse the haproxy-monitor code for inspecting iptables, which means it should be less likely to get out of sync.


pkg/monitor/dynkeepalived.go
```diff
@@ -63,6 +63,32 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
...
 			prevConfig = &newConfig
+
+			// Signal to keepalived whether the haproxy firewall rule is in place
+			ruleExists, err := checkHAProxyPreRoutingRule(apiVip.String(), apiPort, lbPort)
+			if err != nil {
+				log.Error("Failed to check for haproxy firewall rule")
+			} else {
+				filePath := "/var/run/keepalived/iptables-rule-exists"
+				if ruleExists {
+					// FIXME: DRY this
```
cybertron, line 74: I see you. :-P

cmd/dynkeepalived/dynkeepalived.go
```diff
@@ -42,14 +50,16 @@ func main() {
...
+			return monitor.KeepalivedWatch(args[0], clusterConfigPath, args[1], args[2], apiVip, ingressVip, dnsVip, apiPort, lbPort, checkInterval)
 		},
 	}
 	rootCmd.PersistentFlags().StringP("cluster-config", "c", "", "Path to cluster-config ConfigMap to retrieve ControlPlane info")
-	rootCmd.Flags().Duration("check-interval", time.Second*10, "Time between keepalived watch checks")
+	rootCmd.Flags().Duration("check-interval", time.Second*1, "Time between keepalived watch checks")
 	rootCmd.Flags().IP("api-vip", nil, "Virtual IP Address to reach the OpenShift API")
 	rootCmd.PersistentFlags().IP("ingress-vip", nil, "Virtual IP Address to reach the OpenShift Ingress Routers")
 	rootCmd.PersistentFlags().IP("dns-vip", nil, "Virtual IP Address to reach an OpenShift node resolving DNS server")
+	rootCmd.Flags().Uint16("api-port", 6443, "Port where the OpenShift API listens at")
```
bcrochet, line 61: If you have to push another one... Drop the 'at'. :-P Small nit.

pkg/monitor/dynkeepalived.go
bcrochet: Yes please. :D

cmd/dynkeepalived/dynkeepalived.go
cybertron: Ah, these are copy-pasted from the haproxy monitor cli. I'll fix them both.

pkg/monitor/iptables.go
```diff
@@ -55,16 +72,53 @@ func ensureHAProxyPreRoutingRule(apiVip string, apiPort, lbPort uint16) error {
...
+		return nil
+	} else {
+		log.WithFields(logrus.Fields{
+			"spec": strings.Join(ruleSpec, " "),
+		}).Info("Inserting nat OUTPUT rule")
 		return ipt.Insert(table, chain, 1, ruleSpec...)
 	}
 }
+
+func checkHAProxyPreRoutingRule(apiVip string, apiPort, lbPort uint16) (bool, error) {
```
celebdor, line 106: I feel like the name now misrepresents the action since we are checking multiple chain rules, not just prerouting.

cmd/dynkeepalived/dynkeepalived.go
```diff
@@ -42,14 +50,16 @@ func main() {
 				return err
 			}
 
-			return monitor.KeepalivedWatch(args[0], clusterConfigPath, args[1], args[2], apiVip, ingressVip, dnsVip, checkInterval)
+			return monitor.KeepalivedWatch(args[0], clusterConfigPath, args[1], args[2], apiVip, ingressVip, dnsVip, apiPort, lbPort, checkInterval)
 		},
 	}
 	rootCmd.PersistentFlags().StringP("cluster-config", "c", "", "Path to cluster-config ConfigMap to retrieve ControlPlane info")
-	rootCmd.Flags().Duration("check-interval", time.Second*10, "Time between keepalived watch checks")
+	rootCmd.Flags().Duration("check-interval", time.Second*1, "Time between keepalived watch checks")
```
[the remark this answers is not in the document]
celebdor: the only trouble I can think of is to be sending reload commands to keepalived all too often.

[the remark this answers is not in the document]
yboaron: I think that this change will break deployment with unicast-keepalived.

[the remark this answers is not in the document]
yboaron: HAProxy monitor adds the firewall rule after 3*6 Secs and removes it after 2*6 Secs, what's the added value checking rule existence every second?

pkg/monitor/dynkeepalived.go
```diff
@@ -63,6 +63,32 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
 				}
 			}
 			prevConfig = &newConfig
+
+			// Signal to keepalived whether the haproxy firewall rule is in place
+			ruleExists, err := checkHAProxyPreRoutingRule(apiVip.String(), apiPort, lbPort)
+			if err != nil {
+				log.Error("Failed to check for haproxy firewall rule")
+			} else {
+				filePath := "/var/run/keepalived/iptables-rule-exists"
```
yboaron, line 72: I guess it will be better to set filename only once and not every loop iteration

pkg/monitor/dynkeepalived.go
```diff
@@ -63,6 +63,32 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
...
+					// FIXME: DRY this
+					_, err := os.Stat(filePath)
+					if os.IsNotExist(err) {
+						_, err := os.Create(filePath)
+						if err != nil {
+							log.WithFields(logrus.Fields{"path": filePath}).Error("Failed to create file")
+						}
+					}
+				} else {
+					_, err := os.Stat(filePath)
```
yboaron, line 83: Can't we just call to os.remove and consider err=nil and err = file not exist as OK?

pkg/monitor/dynkeepalived.go
yboaron: IIRC, Keepalived monitor runs also in worker nodes, I guess we shouldn't run this check for worker nodes.

Maybe I'm missing something here, but it seems to me that haproxy-monior is the right place to add this logic.

pkg/monitor/iptables.go
cybertron: Valid point. I'll rename.

pkg/monitor/dynkeepalived.go
cybertron: haproxy-monitor doesn't share its /var/run mount with keepalived. I would also argue that haproxy doesn't care about this file. It's consumed by keepalived, so it should be done in the keepalived monitor.

pkg/monitor/dynkeepalived.go
cybertron: Probably, but once I DRY the file existence logic it's one more if statement versus a second clause in another. Plus this way the logic for both creation and removal is the same.

pkg/monitor/dynkeepalived.go
cybertron: Oh, and I moved the filename to a const at the top of the file. Good suggestion.

cmd/dynkeepalived/dynkeepalived.go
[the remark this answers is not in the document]
cybertron: Oops, I missed these comments. The issue with a 10 second check interval is that if haproxy has an issue and needs to remove the firewall rule, it will take 12 seconds for that to happen, and if we're only checking every 10 it may take up to 10 seconds for keepalived to notice that. It nearly doubles the response time to outages (12 to 22 seconds). In the past, keepalived has responded to outages in a couple of seconds at most.

Maybe the longer response time is acceptable. I don't know what our goal for VIP failover is, so I've been trying to keep it as short as possible.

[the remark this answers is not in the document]
yboaron: Well, the monitor sidecars are used for configuration update, and for this purpose, it's reasonable to have slow detection (interval 6-10 Seconds), with this PR, keepalived-monitor sidecar is involved also in failure detection and of course service recovery so it should detect the failure ASAP. 
I know that you moved this code to monitor sidecar because keepalived container doesn't support accessing to iptable information. 

Do you think that adding a new sidecar container to keepalived that will be responsible for this logic is an option?

If we'll decide to keep this logic in keepalived-monitor I think we should have separate timing (and even code/function if it's possible) for failure detection and config update, there is no added value to check for config update every 1 second.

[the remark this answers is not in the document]
cybertron: I put it back to 10 seconds. If it ends up being a problem we can revisit.

cybertron:
/retest

packet failure

pkg/monitor/iptables.go
```diff
@@ -29,42 +31,94 @@ func getProtocolbyIp(ipStr string) iptables.Protocol {
...
+		}
+	}
+	ruleSpec, err = getHAProxyRuleSpec(apiVip, apiPort, lbPort, true)
+	if err != nil {
+		return err
+	}
+	chain = "OUTPUT"
+	if exists, _ := ipt.Exists(table, chain, ruleSpec...); exists {
+		return nil
+	} else {
```
bcrochet, line 98: Is the else necessary here?

pkg/monitor/iptables.go
```diff
@@ -29,42 +31,94 @@ func getProtocolbyIp(ipStr string) iptables.Protocol {
...
 
-	ruleSpec, err := getHAProxyRuleSpec(apiVip, apiPort, lbPort)
+	ruleSpec, err := getHAProxyRuleSpec(apiVip, apiPort, lbPort, false)
 	if err != nil {
 		return err
 	}
+	chain := "PREROUTING"
 	if exists, _ := ipt.Exists(table, chain, ruleSpec...); exists {
 		return nil
 	} else {
```
bcrochet, line 82: Small nit but these else clauses seem superfluous.

pkg/monitor/iptables.go
bcrochet, line 98: Here too.

pkg/monitor/iptables.go
```diff
@@ -29,42 +31,94 @@ func getProtocolbyIp(ipStr string) iptables.Protocol {
...
 	}
 }
+
+func checkHAProxyFirewallRules(apiVip string, apiPort, lbPort uint16) (bool, error) {
+	ipt, err := iptables.NewWithProtocol(getProtocolbyIp(apiVip))
+	if err != nil {
+		return false, err
+	}
+
+	ruleSpec, err := getHAProxyRuleSpec(apiVip, apiPort, lbPort, false)
```
bcrochet, line 112: The magic param at the end doesn't really convey what it is for. One would have to look at the function definition to know, and even then it's not totally clear. Can we change it to maybe add a type for it?

pkg/monitor/iptables.go
cybertron: Done

pkg/monitor/iptables.go
cybertron: Done

cybertron:
/hold cancel

The reasons for the original hold have long since been addressed.

PR #72, opened 2020-07-15
Bug 1857672: Also copy iptables scripts to /usr/sbin

cybertron:
In recent images there are iptables binaries in /usr/sbin/ in
addition to/in place of /usr/bin. This is causing problems because
/usr/sbin is in the PATH before /usr/bin so our code is calling the
wrong iptables. We can just copy the iptables scripts into /usr/sbin
and take care of the problem.


cybertron:
/retest

packet failure

hardys:
/test e2e-metal-ipi

Another packet failure

cybertron:
/test e2e-metal-ipi

Test passed, packet teardown failed. :-(

celebdor:
@cybertron are the new binaries usable?

cybertron:
> @cybertron are the new binaries usable?

The ones being installed in /usr/sbin that I'm overwriting here? No, that's how we found out about them. They were causing the haproxy-monitor container to fail when adding the firewall rule. It was showing up in ci jobs, although I haven't been able to find any recent ones that collected the haproxy-monitor logs.

PR #74, opened 2020-07-23
MGMT-1604 Add VIPS lease to keepalived

YuviGold:
After reading the cmd flags and before executing the keepalived-monitor:
1. Get the directory of args[2] = path_to_config (e.g. /etc/keepalived/keepalived.conf)
2. Check if the there's a file named "monitor.conf" in that directory. If true, then:
	2.1. for [API interface, Ingress interface]:
		2.1.1 Generate a unique MAC address
			2.1.1.1. The first 24 bits of the MAC are fixed of Qumranet prefix.
			2.1.1.2. The last 24 bits are generated using Pearson hashing (clustername + vipname)
		2.1.2. Create an interface type macvlan per vip using the generated mac address
		2.1.3. Run dhclient with the new vip interface in the background
3. Continue regular run


pkg/monitor/lease.go
```diff
@@ -0,0 +1,137 @@
...
+	return macvlanIface.NetInterface(), nil
+}
+
+func generateMac(noise string, iface net.Interface) (mac net.HardwareAddr, err error) {
+	// TODO: Build a more sophisticated mac generation. Keep in mind to depend on clusterName
+	// 1. All the nodes will have the same mac.
+	// 2. Avoid conflicts (with the realIP and the new VIPS)
+	highBits := iface.HardwareAddr.String()[:len(iface.HardwareAddr.String())/2]
+
+	lowBits := strings.Repeat(fmt.Sprintf(":%02x", utils.FletcherChecksum8(noise)+1), 3)
```
cybertron, line 130: I wouldn't use fletcher8 for this. It has a very limited range of values, which leads to collisions. Since it doesn't generate the right number of bits anyway, I'd suggest using something like sha-256 and keeping only the number of bits that you need.

pkg/monitor/lease.go
```diff
@@ -0,0 +1,137 @@
...
+func leaseVIP(name string, vip net.IP) error {
+	iface, err := createVIPInterface(name, vip)
+
+	if err != nil {
+		return err
+	}
+
+	// Renew with dhclient. Maybe can be replaced with https://github.com/insomniacslk/dhcp
+	// TODO: define timeout
+	cmd := exec.Command("dhclient", "-d", "-timeout", "5", iface.Name)
```
cybertron, line 75: This seems problematic. -d is going to keep it in the foreground, which means either this never returns or it returns once and never renews the lease (it's not clear to me which). Since the leaseVIPs call isn't in the monitor loop, I assume the intent was to have dhclient remain running in the background? Otherwise the lease call needs to be moved into the loop. Edit: I see from the comment above that the intent was for this to be run in a goroutine. That isn't happening in the current version so this still needs changes.

Also, we can't actually configure the address on the node since we want keepalived to manage the assignment of the VIP. I believe that means we need -n for this command.

pkg/monitor/lease.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+/*
+	After reading the cmd flags and before executing the keepalived-monitor:
+	1. Get the directory of args[2] = path_to_config (e.g. /etc/keepalived/keepalived.conf) [V]
+	2. Check if the there's a file named "monitor.conf" in that directory. If true, then: [V]
+		2.1. for [API interface, Ingress interface]: . [V]
+			2.1.1 Generate a unique MAC address
+				2.1.1.1. The first 24 bits of the MAC should be the same as the real interface mac address.
```
cybertron, line 23: I don't think we can use the real hardware address as part of the MAC generation. Otherwise the MAC is going to change depending on which system is running the monitor.

pkg/monitor/lease.go
```diff
@@ -0,0 +1,137 @@
...
+						Get the interface by using `GetInterfaceAndNonVIPAddr()` and after that get its mac address.
+				2.1.1.2. The last 24 bits are generated using some hash function.
+						An example is the way the VRIDs are generated in `PopulateVRIDs()`
+			2.1.2. Create an interface type macvlan per vip using the generated mac address [V]
+			2.1.3. Run dhclient with the new vip interface in the background (goroutine)
```
cybertron, line 28: The goroutine part of this seems to be missing.

pkg/monitor/lease.go
celebdor: I recommended using a variable length hash function like https://en.wikipedia.org/wiki/Pearson_hashing to generate all 24 bits at once. I don't think we need a cryptographic hash, @cybertron. I'd suggest one of https://en.wikipedia.org/wiki/List_of_hash_functions#Non-cryptographic_hash_functions .

pkg/monitor/lease.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	return macvlanIface.NetInterface(), nil
+}
+
+func generateMac(noise string, iface net.Interface) (mac net.HardwareAddr, err error) {
+	// TODO: Build a more sophisticated mac generation. Keep in mind to depend on clusterName
+	// 1. All the nodes will have the same mac.
+	// 2. Avoid conflicts (with the realIP and the new VIPS)
+	highBits := iface.HardwareAddr.String()[:len(iface.HardwareAddr.String())/2]
```
celebdor, line 128: As mentioned earlier, please, use `00:1A:4A`

pkg/monitor/lease_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+			_, err = createVIPInterface(testName, testIP)
+			Expect(err).Should(HaveOccurred())
+		})
+
+		AfterEach(func() {
+			Expect(testIface.Name).ShouldNot(Equal(realIface.Name))
+			Expect(testIface.HardwareAddr).ShouldNot(Equal(realIface.HardwareAddr))
+
+			printIPs()
```
celebdor, line 80: What's the purpose of this printing? If we want to test that we created the interface as intended, we should use the net or the netlink golang packages to check it, right?

pkg/monitor/lease.go
cybertron: I was just looking for something that already exists in the Go std lib. :-) Also, I've had issues with security folks for using insecure hashing functions, even for non-security-related functionality.

That said, any hash that gives us at least 24 random-ish bits should be fine here.

pkg/monitor/lease.go
cybertron: Don't we also need -n? We don't want dhclient configuring the VIP on the interface or it may conflict with what keepalived does with it.

Makefile
```diff
@@ -7,7 +7,10 @@ fmt: ## Run go fmt against code
 
 .PHONY: test
 test: ## Run go test against code
-	go test ./pkg/... ./cmd/...
+	docker-compose build $@ && docker-compose run --rm $@ make _$@
```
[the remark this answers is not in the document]
YuviGold: I just found that this repository uses Prow (no CI config files in the repo, had to do the PR to know that), with a src image of Openshift.
You are right it can't run docker-compose at the moment.

Eventually it depends on the image being used in the PR (which can get changed). 
But for now - I will keep the Makefile targets the same. Although I won't be able to add my tests to be a part of the CI. It depends on having `cap_net_admin` capabilities.

pkg/monitor/lease.go
YuviGold: OK :+1:

pkg/monitor/lease_test.go
YuviGold: You are correct. I added it just for self clarification of what's going on as debug printings.

pkg/monitor/lease.go
ronniel1: IMHO Not configuring the interface is handled by the script passed to "-sf"

"-n" does not really work

ronniel1: I think that for now we can just pass `/bin/true`

pkg/monitor/lease.go
```diff
@@ -0,0 +1,137 @@
...
+	// Renew with dhclient. Maybe can be replaced with https://github.com/insomniacslk/dhcp
+	// TODO: define timeout
+	cmd := exec.Command("dhclient", "-d", "-timeout", "5", iface.Name)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	return cmd.Run()
+}
+
+func createVIPInterface(name string, vip net.IP) (*net.Interface, error) {
```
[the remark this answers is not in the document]
ronniel1: This does not really create the interface used for the VIP, right?
It creates the macvlan for the VIP.
The name is kind of misleading

pkg/monitor/lease.go
```diff
@@ -0,0 +1,137 @@
...
+	vipIface, _, err := config.GetInterfaceAndNonVIPAddr([]net.IP{vip})
+	if err != nil {
+		return nil, err
+	}
+
+	mac, err := generateMac(name, vipIface)
```
ronniel1, line 88: mac generation should be a hash on the cluster-name+vip-name.
We need to support multiple clusters on the same network

pkg/monitor/lease.go
YuviGold: No need for a goroutine. executed with `Start()` function instead of `Run()` which execute a goroutine inside.

pkg/monitor/dynkeepalived.go
```diff
@@ -98,10 +98,15 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
 	var appliedConfig, curConfig, prevConfig *config.Node
 	var configChangeCtr uint8 = 0
 
+	clusterName, _, err := config.GetClusterNameAndDomain(kubeconfigPath, clusterConfigPath)
+	if err != nil {
+		return err
+	}
+
 	if exists, err := needLease(cfgPath); err != nil {
```
ronniel1, line 106: Need lease might be on a specific VIP.
User might choose to supply IP for ingress but use dhcp for API

Need to decide if supporting this is critical for now

YuviGold: I suggested to @romfreiman that for now we just use this dumb condition instead of defining how the file would look like.
Do you think it is critical?

pkg/monitor/lease.go
```diff
@@ -0,0 +1,132 @@
...
+	"os/exec"
+	"path/filepath"
+
+	"github.com/milosgajdos/tenus"
+	"github.com/openshift/baremetal-runtimecfg/pkg/config"
+	"github.com/openshift/baremetal-runtimecfg/pkg/utils"
+	"github.com/sirupsen/logrus"
+)
+
+const monitorConf = "monitor.conf"
```
romfreiman, line 16: please rename to unsupported-monitor.conf or something similar

pkg/monitor/dynkeepalived.go
```diff
@@ -98,6 +98,19 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
 	var appliedConfig, curConfig, prevConfig *config.Node
 	var configChangeCtr uint8 = 0
 
+	clusterName, _, err := config.GetClusterNameAndDomain(kubeconfigPath, clusterConfigPath)
+	if err != nil {
+		return err
+	}
+
+	if exists, err := needLease(cfgPath); err != nil {
+		return err
```
romfreiman, line 107: I dont think u should fail - it will break backwards competability. I think no matter what in this stage, u should proceed without dhcp lease and write a log.
@celebdor what do u think

pkg/monitor/lease.go
```diff
@@ -0,0 +1,132 @@
...
+const macPrefixQumranet = "00:1A:4A"
+const leaseInterfaceTemplate = "mc-%s"
+const leaseFile = "/tmp/my.lease"
+
+type VIP struct {
+	name string
+	ip   net.IP
+}
+
+func needLease(cfgPath string) (bool, error) {
```
romfreiman, line 26: please add log

pkg/monitor/lease.go
```diff
@@ -0,0 +1,132 @@
...
+				"vip":  vip,
+			}).Error("Failed to lease a vip")
+			return err
+		}
+	}
+
+	return nil
+}
+
+func leaseVIP(name string, vip net.IP) error {
```
romfreiman, line 55: logs

pkg/monitor/lease.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+	return false, err
+}
+
+func leaseVIPs(clusterName string, vips []VIP) error {
+	for _, vip := range vips {
+		vipFullName := fmt.Sprintf("%s-%s", clusterName, vip.name)
```
romfreiman, line 42: what will be vipFullName? is there any length limit?

pkg/monitor/lease.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func leaseVIP(name string, vip net.IP) error {
+	iface, err := leaseInterface(name, vip)
+
+	if err != nil {
+		return err
+	}
+
+	cmd := exec.Command("dhclient", "-d", "--no-pid", "-sf", "/bin/true", "-lf", leaseFile, "-v", iface.Name, "-H", name)
```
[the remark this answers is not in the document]
celebdor: We should probably monitor the command. If dhclient exits and we "needLease", we should maybe exit as well so that the monitor is restarted.

pkg/monitor/lease.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, err
+	}
+
+	// Check if already exist
+	if iface, err := net.InterfaceByName(fmt.Sprintf(leaseInterfaceTemplate, name)); err == nil {
+		return iface, nil
+	}
+
+	// Create a new network macvlan
+	macvlanIface, err := tenus.NewMacVlanLinkWithOptions(vipIface.Name, tenus.MacVlanOptions{
```
celebdor, line 90: We already vendor a library to interact with the kernel via `netlink`. Please, use the `netlink` library. You can see an example in:
https://github.com/containernetworking/plugins/blob/1fb9793607361bc8ae032a12964c67e7f2b92079/plugins/main/macvlan/macvlan.go#L166
or a simpler veth example:
https://github.com/containernetworking/plugins/blob/master/pkg/ip/link_linux.go#L36

pkg/monitor/lease.go
```diff
@@ -0,0 +1,132 @@
...
+		return iface, nil
+	} else {
+		log.WithFields(logrus.Fields{
+			"interface": macvlanIface.NetInterface().Name,
+		}).Error("Failed to get interface")
+		return nil, err
+	}
+}
+
+func generateMac(noise string, iface net.Interface) (mac net.HardwareAddr, err error) {
```
celebdor, line 123: It feels like generateMac should get the prefix as a parameter and generate as many bytes as necessary to fill what the prefix didn't give.

pkg/utils/utils.go
```diff
@@ -265,3 +265,41 @@ func AddressesRouting(vips []net.IP, af AddressFilter, rf RouteFilter) ([]net.IP
...
+		241, 73, 88, 105, 39, 114, 62, 255, 192, 201, 145, 214, 168, 158, 221,
+		148, 154, 122, 12, 84, 82, 163, 44, 139, 228, 236, 205, 242, 217, 11,
+		187, 146, 159, 64, 86, 239, 195, 42, 106, 198, 118, 112, 184, 172, 87,
+		2, 173, 117, 176, 229, 247, 253, 137, 185, 99, 164, 102, 147, 45, 66,
+		231, 52, 141, 211, 194, 206, 246, 238, 56, 110, 78, 248, 63, 240, 189,
+		93, 92, 51, 53, 183, 19, 171, 72, 50, 33, 104, 101, 69, 8, 252, 83, 120,
+		76, 135, 85, 54, 202, 125, 188, 213, 96, 235, 136, 208, 162, 129, 190,
+		132, 156, 38, 47, 1, 7, 254, 24, 4, 216, 131, 89, 21, 28, 133, 37, 153,
+		149, 80, 170, 68, 6, 169, 234, 151,
+	}
```
celebdor, line 289: How did you generate this particular table?

README.md
```diff
@@ -19,3 +19,14 @@ The available flags are:
 * --ingress-vip Virtual IP Address to reach the OpenShift Ingress Routers
 
 Note that you must pass at least one VIP for the VRRP interface to be found.
+
+## Test
+
+In order to run the tests you should have all these prerequisites:
+* make
+* docker
+* docker-compose
```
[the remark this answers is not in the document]
YuviGold: 1. I will extend the README to be clearer.
2. Added a capability to check to the tests so it skips them on regular test w/o capabilities.
```golang
		if !hasCap(capability.CAP_NET_ADMIN, capability.CAP_NET_RAW) {
			Skip("Must run with capabilities: CAP_NET_ADMIN, CAP_NET_RAW")
		}
```
3. Some options for the CI:
* openshift/release support Docker-in-Docker. You can change the `src` container for `unit` to include `docker-compose` and run `docker_test`
* Run on a packet server with capbilities
* Run the `src` container with capabilities and run the same `test` target. I'm not sure if it's possible today on Prow.

pkg/monitor/lease.go
YuviGold: Good point. Fixed
```golang
const interfaceMaxSize = 12

		vipFullName := fmt.Sprintf("%s-%s", clusterName, vip.name)
		// Takes the last `interfaceMaxSize` bytes of vipFullName
		vipFullName = vipFullName[int(math.Max(0, float64(len(vipFullName)-interfaceMaxSize))):]
```

pkg/monitor/lease.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/milosgajdos/tenus"
+	"github.com/openshift/baremetal-runtimecfg/pkg/config"
+	"github.com/openshift/baremetal-runtimecfg/pkg/utils"
+	"github.com/sirupsen/logrus"
+)
+
+const monitorConf = "monitor.conf"
+const macPrefixQumranet = "00:1A:4A"
+const leaseInterfaceTemplate = "mc-%s"
+const leaseFile = "/tmp/my.lease"
```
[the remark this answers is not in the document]
YuviGold: File per vip isn't necessary because the lease file can be appended. But done anyway for clarity.

pkg/utils/utils.go
YuviGold: It is written in the comment - `RFC 3074` under `Hash Function for Load Balancing`

pkg/monitor/dynkeepalived.go
YuviGold: Will be defined later.

pkg/monitor/lease.go
```diff
@@ -0,0 +1,179 @@
...
+		log.WithFields(logrus.Fields{
+			"name": name,
+		}).WithError(err).Error("Failed to read new device")
+		return nil, err
+	}
+
+	return macVlanIfc, nil
+}
+
+func generateMac(prefix []byte, noise string) (mac net.HardwareAddr) {
```
cybertron, line 161: Nit: I wouldn't call this "noise". That suggests a randomness that may or may not be wanted. It's just a key passed to a hash function, so I would call it "key".

pkg/monitor/lease_test.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func getOutput(command string) string {
+	cmd := exec.Command("bash", "-c", command)
+	out, err := cmd.Output()
+	if err != nil {
+		log.Fatal(command, ": ", err)
+	}
+
+	// fmt.Println(cmd.Args)
```
cybertron, line 272: Nit: Leftover comment

stbenjam:
We don't typically reference jira issues on GitHub

pkg/config/node.go
```diff
@@ -419,3 +415,14 @@ func GetLBConfig(kubeconfigPath string, apiPort, lbPort, statPort uint16, apiVip
 	}).Debug("Config for LB configuration retrieved")
 	return config, nil
 }
+
+func GetClusterNameAndDomain(kubeconfigPath, clusterConfigPath string) (clusterName string, clusterDomain string, err error) {
+	// Try cluster-config.yml first
+	clusterName, clusterDomain, err = getClusterConfigClusterNameAndDomain(clusterConfigPath)
+	if err != nil {
+		// We are using kubeconfig as a fallback for this
+		clusterName, clusterDomain, err = GetKubeconfigClusterNameAndDomain(kubeconfigPath)
```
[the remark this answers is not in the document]
YuviGold: I copied the same behavior into a function. 
This function always returns the 3 parameters - `clusterName, clusterDomain, err` . 
Whoever calls it this function, should check `err` value and decide to do w/e he wants - e.g. log it

pkg/monitor/dynkeepalived.go
```diff
@@ -98,6 +98,27 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
 	var appliedConfig, curConfig, prevConfig *config.Node
 	var configChangeCtr uint8 = 0
 
+	// Lease VIPS
+	// To avoid backwards compatability - Bad flow doesn't return an error
+	if clusterName, _, err := config.GetClusterNameAndDomain(kubeconfigPath, clusterConfigPath); err == nil {
+		if exists, err := needLease(cfgPath); err == nil && exists {
+			vips := []VIP{{"api", apiVip}, {"ingress", ingressVip}}
+			if err = leaseVIPs(cfgPath, clusterName, vips); err == nil {
```
ori-amizur, line 106: In case that the leaseVIPs fails, what is the meaning of it in terms of successful completion of the installation?  Maybe VIPs are not available and we will not indicate it, only in the logs.

pkg/monitor/lease.go
```diff
@@ -0,0 +1,179 @@
...
+			"masterDevice": masterDevice,
+			"name":         name,
+		}).WithError(err).Error("Failed to lease interface")
+		return err
+	}
+
+	cmd := exec.Command("dhclient", "-d", "--no-pid", "-sf", "/bin/true",
+		"-lf", getLeaseFile(cfgPath, name), "-v", iface.Name, "-H", name)
+
+	return cmd.Start()
```
ori-amizur, line 92: Why don't we monitor the process? How can we know if the lease succeeded?

pkg/monitor/dynkeepalived.go
```diff
@@ -98,6 +98,27 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
...
+			} else {
+				log.WithFields(logrus.Fields{
+					"cfgPath":     cfgPath,
+					"clusterName": clusterName,
+					"vips":        vips,
+				}).WithError(err).Error("Failed to lease VIPS")
+			}
+		}
+	}
+
```
ori-amizur, line 121: Shouldn't we monitor the lease files to see if we got addresses, and at least log if we did not?

pkg/monitor/dynkeepalived.go
YuviGold: Fixed.

pkg/config/node.go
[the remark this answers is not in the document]
cybertron: Ah, right. We're not doing anything with these values in this function anymore.

PR #76, opened 2020-07-28
TEST: install dhcp-client

cybertron:


cybertron:
/hold

Just trying to figure out what will work in ci for this.

cybertron:
I missed that this was being handled in https://github.com/openshift/release/pull/10498 Shouldn't need this then.

PR #79, opened 2020-08-07
Try to read first from localhost and failover to api-vip

yboaron:


yboaron:
Tried some ideas for handling CI deployment problem  - all of them didn't help.
/close

PR #80, opened 2020-08-07
Check if iptable rule exist only for master nodes

yboaron:


yboaron:
Didn't help with CI failure
/close

PR #82, opened 2020-08-09
Retrieve master details for bootstrap also from client info

yboaron:


pkg/config/node.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 const localhostKubeApiServerUrl string = "https://localhost:6443"
-const bootstrapIpServerPort string = "64444"
+const bootstrapIpServerPort string = "64445"
```
celebdor, line 27: Why do we change the port number?

cmd/dynkeepalived/dynkeepalived.go
```diff
@@ -54,7 +54,7 @@ func main() {
 		},
 	}
 	rootCmd.PersistentFlags().StringP("cluster-config", "c", "", "Path to cluster-config ConfigMap to retrieve ControlPlane info")
-	rootCmd.Flags().Duration("check-interval", time.Second*10, "Time between keepalived watch checks")
+	rootCmd.Flags().Duration("check-interval", time.Second*15, "Time between keepalived watch checks")
```
celebdor, line 57: This is just reducing the race chance, right? Not addressing the source

pkg/monitor/dynkeepalived.go
```diff
@@ -70,6 +71,26 @@ func updateUnicastConfig(kubeconfigPath string, newConfig, appliedConfig *config
...
+
+		for _, value := range serverBackends {
+			ipExist := false
+			for _, lbValue := range newConfig.LBConfig.Backends {
+				if value.Address == lbValue.Address {
+					ipExist = true
+				}
+			}
+			if !ipExist {
+				newConfig.LBConfig.Backends = append(newConfig.LBConfig.Backends, config.Backend{Host: "", Address: value.Address})
```
celebdor, line 84: Instead of managing ipExist, isn't it the same to put this block inside line 80 (instead of ipExist=True) ?

pkg/config/node.go
yboaron: We don't need, I just wanted to run CI  without sending MCO PR for deleting the unicast-server container in bootstrap manifest

cmd/dynkeepalived/dynkeepalived.go
yboaron: Yes and Not,
Well, In case bootstrap  failed to read master nodes details from OCP API , we have the following flow:

1. Master-x send a request to the unicast server at bootstrap node 
2. The unicast server at the bootstrap should : 
  2.1 add master-x IP address to an array (stores the clients/masters IPs learned by the server)
  2.2 Reply with his IP address
3. Master-x will add bootstrapIP to its configuration, and after debouncer of 3 ( poll interval ) , the new config will be written to keepalived config file 
4. From the bootstrap side, In the next poll interval the bootstrap will include master-x IP in his configuration ( same here 3*toll interval)

So, in the worst case, master-x will have 2 poll interval (2*10) before it adds the bootstrap IP, and bootstrap has 3 poll intervals (3*5 seconds) 
I guess we can delete this fix,  I wanted to see green CI first

pkg/config/node.go
yboaron: so, to avoid port conflict, I changed the port value here

pkg/monitor/dynkeepalived.go
yboaron: Hmm, but we need to add it to the backends only if it doesn't exist.
Only after we scan all the array and didn't find it.

yboaron:
Since our plan is to remove the unicast server, we should first look for other alternatives to reduce the race during deployment.

PR #84, opened 2020-08-10
Bug 1867080 : Keepalived monitor - store correctly current configuration

yboaron:
Keepalived monitor reads every x seconds the current configuration, it compares it to the previous configuration. In case a change is detected a new Keepalived conf file is rendered. 

In one of the latest commits the variable used to store the current config changed by mistake to be local variable instead of runtime-created variable (e.g: newConfig  := some_value ).

This PR fixes this mistake.


yboaron:
/ retitle Bug 1867080  : Keepalived monitor - store correctly current configuration

yboaron:
/retitle Bug 1867080 : Keepalived monitor - store correctly current configuration

PR #85, opened 2020-08-12
Install diffutils in image

cybertron:
This is required for the haproxy-monitor liveness probe. It apparently
got dropped when we switched to the ubi8 image.


cybertron:
/retest

Looks like packet setup hung.

PR #87, opened 2020-08-17
Read BootstrapIP from OCP API

yboaron:
In unicast Keeplaived, master nodes should register the bootstrap as one of the peers.
With this PR master nodes retrieve the Bootstrap IP address from OCP API instead of
reading it from a dedicated server that runs on the bootstrap node.

A follow-up PRs should delete the unicast server capability in both baremetal-runtimecfg and the MCO.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	golang.org/x/sync v0.0.0-20190423024810-112230192c58 // indirect
 	golang.org/x/sys v0.0.0-20190616124812-15dcb6c0061f
 	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4 // indirect
-	gopkg.in/yaml.v2 v2.2.4 // indirect
+	gopkg.in/yaml.v2 v2.2.4
 	k8s.io/api v0.0.0-20190712022805-31fe033ae6f9
 	k8s.io/apimachinery v0.0.0-20190711222657-391ed67afa7b
 	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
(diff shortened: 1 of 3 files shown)


cybertron:
This looks reasonable and passed CI so I assume it's working. The one concern I have is that there are references to things owned by other components that could conceivably change and break us. I'm thinking things like host-etcd-2 (is 2 a version or does it have some other meaning?) and alpha.installer.openshift.io/etcd-bootstrap, which presumably being alpha is subject to change.

Have we discussed this with the etcd team to determine whether we should rely on these things being present? I don't know how likely it is that they would change, but this seems like the sort of thing that has broken us in the past.

yboaron:
> This looks reasonable and passed CI so I assume it's working. The one concern I have is that there are references to things owned by other components that could conceivably change and break us. I'm thinking things like host-etcd-2 (is 2 a version or does it have some other meaning?) and alpha.installer.openshift.io/etcd-bootstrap, which presumably being alpha is subject to change.
> 
> Have we discussed this with the etcd team to determine whether we should rely on these things being present? I don't know how likely it is that they would change, but this seems like the sort of thing that has broken us in the past.

Agree, that's a good point.
The 'cleaner' solution would be to retrieve the bootstrap IP in bootkube.sh and pass it as a parameter to MCO, though this kind of solution isn't feasible for 4.6.

I'll double-check this point with etcd folks .

Thanks

yboaron:
/hold
Until we'll decide if it's safe to rely on etcd-endpoints annotation

yboaron:
Masters don't need to know about the bootstrap IP anymore so we can close this PR

/close

PR #89, opened 2020-08-17
Bug 1866347: monitoring keepalive wrong format

YuviGold:
* Receiving mac address from the monitoring configuration file in yaml.
Deploying multiple assisted installers where the user does not provide the VIP addresses with the same cluster name ends up with them competing for the same DHCP leased address and conflicting.
The format should be as follows:
```yaml
- name: api
  mac-address: ae:3e:38:f5:f8:15
  ip-address: 1.2.3.4
- name: ingress
  mac-address: 80:32:53:4f:cf:d6
  ip-address: 1.2.3.5

```

* Watching the lease file with an `inotify` implementation https://github.com/fsnotify/fsnotify 
In order to verify the `dhclient` made a renew for the lease instead of creating a new one.


cybertron:
Can you provide more details in the description on why these changes are being made?

pkg/monitor/lease.go
```diff
@@ -2,71 +2,95 @@ package monitor
...
+		mac, err := net.ParseMAC(vip.MacAddress)
 
-		vipMasterIface, _, err := config.GetInterfaceAndNonVIPAddr([]net.IP{vip.ip})
 		if err != nil {
 			log.WithFields(logrus.Fields{
-				"name": vip.name,
-				"ip":   vip.ip,
-			}).WithError(err).Error("Failed to get the master device for a vip")
+				"vip": vip,
+			}).WithError(err).Error("Failed to parse mac")
```
romfreiman, line 84: assuming u have 2 and one failed, what will happen? one will get lease and the other won't? If so, u have assymetry here, right? depends which on will fail first.
Not sure what the right behavior but at least I would align it - either best effort for both, or neither.

YuviGold: In case anything in the leasing process would fail - the program fails as well.

YuviGold:
> Can you provide more details in the description on why these changes are being made?

Added. Full description can be found in [Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1866347)

romfreiman:
I think it worth mentioning in the PR the format of the expected file

pkg/monitor/lease.go
```diff
@@ -2,71 +2,95 @@ package monitor
...
-	name string
-	ip   net.IP
+type vip struct {
+	Name       string `yaml:"name"`
+	MacAddress string `yaml:"mac-address"`
+	IpAddress  string `yaml:"ip-address"`
 }
 
-func needLease(cfgPath string) (bool, error) {
+func getVipsToLease(cfgPath string) (vips *[]vip, err error) {
```
ori-amizur, line 28: No need to return pointer to slice.  Slice can be nil too.

pkg/monitor/lease.go
```diff
@@ -86,16 +110,139 @@ func leaseVIP(cfgPath, masterDevice, name string) error {
...
+			"filename":      lease_file,
+			"expectedIface": iface.Name,
+			"expectedIp":    ip,
+		}).WithError(err).Error("Failed to watch the lease file")
+		return err
+	}
+
 	cmd := exec.Command("dhclient", "-d", "--no-pid", "-sf", "/bin/true",
-		"-lf", getLeaseFile(cfgPath, name), "-v", iface.Name, "-H", name)
+		"-lf", lease_file, "-v", iface.Name, "-H", name)
```
ori-amizur, line 133: We are watching the file, but we will not know if the dhclient terminated for some reason.  Shouldn't we monitor the process as well and maybe restart if it terminates?

pkg/monitor/lease.go
```diff
@@ -86,16 +110,139 @@ func leaseVIP(cfgPath, masterDevice, name string) error {
...
+							"filename": fileName,
+						}).WithError(err).Error("Failed to get lease information from leasing file")
+					} else if iface != expectedIface || ip != expectedIp {
+						log.WithFields(logrus.Fields{
+							"filename":      fileName,
+							"iface":         iface,
+							"expectedIface": expectedIface,
+							"ip":            ip,
+							"expectedIp":    expectedIp,
+						}).WithError(err).Error("A new lease has been written to the lease file with wrong data")
```
ori-amizur, line 174: Maybe we should add the new ip in the log message?

YuviGold: It is part of the fields.

cybertron:
/retitle Bug 1866347: monitoring keepalive wrong format

cybertron:
Is this even a valid use case? Duplicate cluster names in a single DHCP domain are going to cause other issues besides this. Shouldn't we avoid that in the first place?

YuviGold:
> Is this even a valid use case? Duplicate cluster names in a single DHCP domain are going to cause other issues besides this. Shouldn't we avoid that in the first place?

@cybertron 
In assisted-service - Users can create multiple clusters, and we do not know which of them will be connected to the same DHCP server, it will be difficult to make sure that we don't have duplicate names. Hence we validate the clusterID.

bcrochet:
Did all of the new vendoring really need to happen? Is that central to the fix?

cybertron:
/retest

Failed on an infra issue.

pkg/monitor/lease.go
```diff
@@ -86,16 +122,204 @@ func leaseVIP(cfgPath, masterDevice, name string) error {
...
+			"filename": leaseFile,
+		}).WithError(err).Error("Failed to create a watcher for lease file")
+		return err
+	}
+
+	// -sf avoiding dhclient from setting the received IP to the interface
+	// --no-pid in order to allow running multiple dhclients
+	commandArgs := []string{"-v", iface.Name, "-H", name,
+		"-sf", "/bin/true", "-lf", leaseFile, "-d",
+		"--no-pid", "-pf", fmt.Sprintf("/var/run/dhclient.%s.pid", iface.Name)}
```
cybertron, line 148: It doesn't make sense to pass both --no-pid and -pf. I'd be surprised if this does what you want.

pkg/monitor/lease.go
```diff
@@ -86,16 +122,204 @@ func leaseVIP(cfgPath, masterDevice, name string) error {
...
+
+	cmd := exec.Command("dhclient", commandArgs...)
 	cmd.Stderr = os.Stderr
 
-	return cmd.Start()
+	if !runInfinite {
```
cybertron, line 153: Why is this configurable? AFAICT we only ever call this with true. It seems like a whole bunch of dead code.

pkg/monitor/lease.go
```diff
@@ -86,16 +122,204 @@ func leaseVIP(cfgPath, masterDevice, name string) error {
...
+		return cmd.Process.Kill()
+	} else {
+		RunInfiniteWatcher(log, watcher, leaseFile, iface.Name, ip)
+		return cmd.Start()
+	}
 }
 
-func leaseInterface(masterDevice string, name string) (*net.Interface, error) {
-	mac := GenerateMac(MacPrefixQumranet[:], name)
+func createLeaseFileWatcher(log logrus.FieldLogger, fileName string) (*fsnotify.Watcher, error) {
```
cybertron, line 177: Ditto my earlier comment about passing in a logger unnecessarily.

pkg/monitor/lease.go
```diff
@@ -86,16 +122,204 @@ func leaseVIP(cfgPath, masterDevice, name string) error {
...
+
+	cmd := exec.Command("dhclient", commandArgs...)
 	cmd.Stderr = os.Stderr
 
-	return cmd.Start()
+	if !runInfinite {
+		write := make(chan error)
+		defer close(write)
+
+		RunFiniteWatcher(log, watcher, leaseFile, iface.Name, ip, write)
```
ori-amizur, line 157: I think that if you run with -1 flag, then it will run once, and then it will terminate.  After that the lease file can be checked.  Seems simpler flow.

pkg/monitor/lease.go
```diff
@@ -86,16 +122,204 @@ func leaseVIP(cfgPath, masterDevice, name string) error {
...
+			"filename": fileName,
+		}).WithError(err).Error("Lease file watcher error")
+		return true, err
+	}
+
+	return false, nil
+}
+
+func GetLastLeaseFromFile(log logrus.FieldLogger, fileName string) (string, string, error) {
+	data, err := ioutil.ReadFile(fileName)
```
ori-amizur, line 276: I think the only common functions should be GetLastLeaseFromFile, and LeaseInterface.  Otherwise, it will make both implementations more complicated that if implemented separately.

pkg/monitor/lease.go
```diff
@@ -2,71 +2,107 @@ package monitor
...
+		return nil, err
+	}
+
+	log.Info("Valid file format")
+
+	return &vips, nil
 }
 
-func leaseVIPs(cfgPath string, clusterName string, vips []VIP) error {
+func LeaseVIPs(log logrus.FieldLogger, cfgPath string, clusterName string, vipMasterIface string, vips []vip, runInfinite bool) error {
```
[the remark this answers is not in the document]
YuviGold: It can use and it actually uses the same one when it is called from `dynkeepalived.go`
```golang
if err = LeaseVIPs(log, cfgPath, vipIface.Name, []vip{*vips.APIVip, *vips.IngressVip}, true); err != nil {
```

We want to use the leasing functions from the `monitor` package in the `assisted-service` as well.
You define your own log in the monitor package via `var log = logrus.New()`.
In case another package with their own logger wants to use it (or even for it to be manipulated for the tests) it would be impossible without giving a log parameter.
Hence a log parameter is more flexible

pkg/monitor/lease.go
YuviGold: It could be much simpler but unfortunately couldn't make it work

pkg/monitor/lease.go
YuviGold: * `--no-pid` in order to allow running multiple `dhclient` simultaneously
* `-pf` allow killing the process

Couldn't make it work without them both

pkg/monitor/lease.go
YuviGold: I agree.

pkg/utils/fswatcher.go
```diff
@@ -0,0 +1,60 @@
...
+	"os"
+
+	"github.com/sirupsen/logrus"
+	"gopkg.in/fsnotify.v1"
+)
+
+func CreateFileWatcher(log logrus.FieldLogger, fileName string) (*fsnotify.Watcher, error) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		log.WithError(err).Error("Failed to add a create a new watcher")
```
cybertron, line 13: Nit: I think "add a" should be removed from this log message.

pkg/monitor/lease.go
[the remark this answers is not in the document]
cybertron: In the future, when adding a new use case for code like this it would be a good thing to highlight in the PR description. That's important context for reviewers.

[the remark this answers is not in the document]
bcrochet: Personally, I would add new parameters to the end of the signature, not the beginning. Also, I'm not a huge fan of bare booleans. There is no context at the caller as to what a magic "true" or "false" might mean.

PR #91, opened 2020-08-27
Bug 1871769: Keepalived-monitor: delete bootstrap from peers list

yboaron:
Delete bootstrap from API-VIP peers list after bootstrap complete phase.

### DIFF
--- a/pkg/monitor/dynkeepalived.go
+++ b/pkg/monitor/dynkeepalived.go
@@ -30,7 +30,8 @@ const (
 )
 
 var (
-	gBootstrapIP string
+	gBootstrapIP     string
+	gUnicastServerIP string
 )
 
 func getActualMode(cfgPath string) (error, bool) {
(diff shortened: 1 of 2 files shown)


yboaron:
/retitle bug 1871769 :  Keepalived-monitor: delete bootstrap from peers list

pkg/monitor/dynkeepalived.go
```diff
@@ -30,7 +30,7 @@ const (
 )
 
 var (
-	gBootstrapIP string
+	gBootstrapIP string = ""
```
cybertron, line 33: Is this necessary? Strings already default to "".

pkg/monitor/dynkeepalived.go
```diff
@@ -87,20 +87,22 @@ func doesConfigChanged(curConfig, appliedConfig *config.Node) bool {
...
 		return
 	}
-	gBootstrapIP, err = config.GetBootstrapIP(apiVip)
-	if err != nil {
-		log.Debugf("Could not retrieve bootstrap IP: %v", err)
+
+	retrievedIP, err := config.GetBootstrapIP(apiVip)
+	if gBootstrapIP != retrievedIP {
+		if err != nil {
+			log.Infof("Could not retrieve bootstrap IP: %v , retrieved IP: %s", err, retrievedIP)
```
cybertron, line 100: I wonder if this should be debug. Isn't it going to happen every time the monitor runs after the bootstrap goes down?

Also a nit: retrievedIP will always be "" here. I don't think we need to include it.

pkg/monitor/dynkeepalived.go
```diff
@@ -87,20 +87,22 @@ func doesConfigChanged(curConfig, appliedConfig *config.Node) bool {
...
 	if os.Getenv("IS_BOOTSTRAP") == "yes" {
 		gBootstrapIP = ""
 		return
 	}
-	gBootstrapIP, err = config.GetBootstrapIP(apiVip)
-	if err != nil {
-		log.Debugf("Could not retrieve bootstrap IP: %v", err)
+
+	retrievedIP, err := config.GetBootstrapIP(apiVip)
+	if gBootstrapIP != retrievedIP {
```
cybertron, line 98: Feels a little weird to be using the return value when err != nil, but I guess it's fine since GetBootstrapIP returns "" on error.

pkg/monitor/dynkeepalived.go
yboaron: Hmm, this log should happen only when the current bootstrap IP (retrievedIP) is different from the previous bootstrap IP ( gBootstrapIP), so I assume we should see this line twice on the logs.

pkg/monitor/dynkeepalived.go
yboaron: Done

yboaron:
/hold 
Until I'll check why local deployment fails

yboaron:
/hold cancel

I understood the problem, bootstrap should be deleted from peers list **only** if bootstrap node doesn't reply to client requests.
I fixed that.

celebdor:
/retitle Bug 1871769: Keepalived-monitor: delete bootstrap from peers list

yboaron:
With [1]   masters don't need to know about the bootstrap, so after [1] will be merged we can close this PR 

[1] https://github.com/openshift/machine-config-operator/pull/2107

PR #92, opened 2020-08-30
Bug 1866265: Stop Keepalived on bootstrap after bootstrap completed

yboaron:
When OPENSHIFT_INSTALL_PRESERVE_BOOTSTRAP deployment configuration parameter is set to true the
bootstrap node won't be destroyed and the infrastructure components including Keepalived on the bootstrap continue to run.

With this PR a stop message is sent to Keepalived control socket when the API server is not operational anymore on bootstrap node.
This change is important for platforms that run Keepalived in unicast mode (baremetal).

### DIFF
--- a/pkg/monitor/dynkeepalived.go
+++ b/pkg/monitor/dynkeepalived.go
@@ -19,15 +19,16 @@ import (
 )
 
 const (
-	keepalivedControlSock                      = "/var/run/keepalived/keepalived.sock"
-	cfgKeepalivedChangeThreshold uint8         = 3
-	dummyPortNum                 uint16        = 123
-	unicastPatternInCfgFile                    = "unicast_peer"
-	modeUpdateFilepath                         = "/etc/keepalived/monitor.conf"
-	userModeUpdateFilepath                     = "/etc/keepalived/monitor-user.conf"
-	modeUpdateIntervalInSec      time.Duration = 600
-	processingTimeInSec          uint16        = 30
-	iptablesFilePath                           = "/var/run/keepalived/iptables-rule-exists"
+	keepalivedControlSock                       = "/var/run/keepalived/keepalived.sock"
+	cfgKeepalivedChangeThreshold  uint8         = 3
+	dummyPortNum                  uint16        = 123
+	unicastPatternInCfgFile                     = "unicast_peer"
+	modeUpdateFilepath                          = "/etc/keepalived/monitor.conf"
+	userModeUpdateFilepath                      = "/etc/keepalived/monitor-user.conf"
+	modeUpdateIntervalInSec       time.Duration = 600
+	processingTimeInSec           uint16        = 30
+	iptablesFilePath                            = "/var/run/keepalived/iptables-rule-exists"
+	bootstrapApiFailuresThreshold int           = 4
 )
 
 var (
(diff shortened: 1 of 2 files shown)


yboaron:
/retitle bug: 1866265 Stop Keepalived on bootstrap after bootstrap completed

pkg/monitor/dynkeepalived.go
```diff
@@ -141,6 +142,32 @@ func isModeUpdateNeeded(cfgPath string) (bool, modeUpdateInfo) {
 	return updateRequired, desiredModeInfo
 }
 
+func handleBootstrapApiNA(kubeconfigPath string, bootstrapApiNA chan bool) {
```
bcrochet, line 145: Lots of extra lines here.

bcrochet, line 145: It's not immediately apparent what "NA" stands for... It's used a few times here, and I sort of get that it's possibly an error condition, but I'm not totally sure from the context. Can we be more explicit here?

pkg/monitor/dynkeepalived.go
```diff
@@ -235,6 +263,10 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
 
 	go handleConfigModeUpdate(cfgPath, kubeconfigPath, updateModeCh)
 
+	if os.Getenv("IS_BOOTSTRAP") == "yes" {
```
bcrochet, line 266: I really prefer the os.LookupEnv() method. It returns true or false based on existence. If it isn't set, it returns false. If it's set, it returns true. That way, we don't really care what the value is, just that it's set. IIUC, we only set IS_BOOTSTRAP on the bootstrap, right?

pkg/monitor/dynkeepalived.go
```diff
@@ -141,6 +142,32 @@ func isModeUpdateNeeded(cfgPath string) (bool, modeUpdateInfo) {
 	return updateRequired, desiredModeInfo
 }
 
+func handleBootstrapApiNA(kubeconfigPath string, bootstrapApiNA chan bool) {
+	consecutiveErr := 0
+
+	for {
+
```
bcrochet, line 149: Extra line.

pkg/monitor/dynkeepalived.go
```diff
@@ -244,6 +276,19 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
 		select {
 		case <-done:
 			return nil
```
bcrochet, line 278: An extra line here would be nice. :)

bcrochet (commented):
Couple of comments, couple of questions.

pkg/monitor/dynkeepalived.go
yboaron: Nope, we set value ("yes" for bootstrap and "no" for rest of the nodes) see [1], if we'll change the way we set these env variables it will make sense to use os.LookupEnv() . 

I'm not sure we can change this thing at this point (code freeze)  


[1] https://github.com/openshift/machine-config-operator/blob/master/templates/common/baremetal/files/baremetal-keepalived.yaml#L128-#129

pkg/monitor/dynkeepalived.go
yboaron: Changed the name

pkg/monitor/dynkeepalived.go
bcrochet: Fair enough.

bcrochet (approved):

yboaron:
retest

bcrochet (commented):
/lgtm

celebdor:
/retitle  Bug 1866265: Stop Keepalived on bootstrap after bootstrap completed

bcrochet (commented):
/lgtm

PR #93, opened 2020-09-01
Bug 1872080: Add Dockerfile.openshift for CI

bcrochet:
We want to use this Dockerfile for CI so that local builds will continue
to use the ubi8 image.

### DIFF
--- a/Dockerfile.openshift
+++ b/Dockerfile.openshift
@@ -1,0 +1,26 @@
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-openshift-4.6 AS builder
+WORKDIR /go/src/github.com/openshift/baremetal-runtimecfg
+COPY . .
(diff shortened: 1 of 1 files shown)


cybertron:
So this would be used for ci and product builds as defined by https://github.com/openshift/ocp-build-data/blob/openshift-4.6/images/baremetal-runtimecfg.yml right? Should this be called Dockerfile.openshift instead, since we're already using a rhel8-based image in the regular Dockerfile?

bcrochet:
> So this would be used for ci and product builds as defined by https://github.com/openshift/ocp-build-data/blob/openshift-4.6/images/baremetal-runtimecfg.yml right? Should this be called Dockerfile.openshift instead, since we're already using a rhel8-based image in the regular Dockerfile?

Don't care what it's called. Dockerfile.openshift does work better for me.

cybertron:
/lgtm

Looks like we will need a bug for this to merge.

bcrochet:
/retitle Bug 1872080: Add Dockerfile.openshift for CI

PR #95, opened 2020-09-03
Change local Dockerfile to use CentOS 8

bcrochet:
Now that CI and Productization uses Dockerfile.openshift, the "main"
Dockerfile is free to use CentOS 8. This should limit the drift.


cybertron:
/lgtm

4.7 is open, let's get this in so I stop having bogus image build failures locally. :-)

PR #100, opened 2020-09-17
Change keepalived VIP subnet mask to host mask

yboaron:
To work around [1]  bug in network manager the subnet mask of Keepalived VIP address was set to the value of control plane IP subnet mask.
So, after the network manager bug was resolved we can safely change VIP subnet mask to a host subnet mask.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1700415


yboaron:
/retitle Change keepalived VIP subnet mask to host mask

PR #101, opened 2020-09-26
Bug 1872632: runtimecfg node-ip: allow running without VIPs

danwinship:
"runtimecfg node-ip" is currently only used on bare metal IPI, and depends on being given an apiserver VIP to use as a hint to figure out the node IP to use. In some cases kubelet also needs help figuring out the default node IP on bare metal UPI, but we have no apiserver VIP to use there. This extends "runtimecfg node-ip" to allow running with no VIPs, in which case it just picks an IP based on the default route.

/cc @celebdor


cybertron:
/lgtm

I played around with this locally and it handled all of the combinations of addresses that I tried. Should be good to go.

PR #102, opened 2020-10-01
Bug 1884420: bootstrap: API shows up, start it again

celebdor:
In order to keep the VIP in the bootstrap node until the masters' API shows up, we increased the priority of the bootstrap keepalived API VIP membership. In order for the VIP to successfully move to the masters even when the bootstrap is requested to stay even after clustering (when its API server is already gone), we implemented a mechanism in the monitor that stops it. The problem with that was that sometimes, during a clustering, the API in the bootstrap node could go down for long enough that it looked like it would not go up anymore.

This PR addresses it by continuing to check for the API server on the bootstrap node, and reloading keepalived if it shows up again. In case it is gone for good, the behavior will be the same, but if it just went down for a while because of API pod restarts and resource issues, we'll reload and reclaim the API VIP.

### DIFF
--- a/pkg/monitor/dynkeepalived.go
+++ b/pkg/monitor/dynkeepalived.go
@@ -31,6 +31,13 @@ const (
 	bootstrapApiFailuresThreshold int           = 4
 )
 
+type APIState uint8
+
+const (
+	stopped APIState = iota
+	started APIState = iota
+)
+
 var (
 	gBootstrapIP string
 )
@@ -143,7 +150,7 @@ func isModeUpdateNeeded(cfgPath string) (bool, modeUpdateInfo) {
 	return updateRequired, desiredModeInfo
 }
 
-func handleBootstrapStopKeepalived(kubeconfigPath string, bootstrapStopKeepalived chan bool) {
+func handleBootstrapStopKeepalived(kubeconfigPath string, bootstrapStopKeepalived chan APIState) {
 	consecutiveErr := 0
 
 	/* It could take up to ~20 seconds for the local kube-apiserver to start running on the bootstrap node,
@@ -166,6 +173,9 @@ func handleBootstrapStopKeepalived(kubeconfigPath string, bootstrapStopKeepalive
 				"consecutiveErr": consecutiveErr,
 			}).Info("handleBootstrapStopKeepalived: detect failure on API")
 		} else {
+			if consecutiveErr > bootstrapApiFailuresThreshold { // Means it was stopped
+				bootstrapStopKeepalived <- started
+			}
 			consecutiveErr = 0
 		}
 		if consecutiveErr > bootstrapApiFailuresThreshold {
@@ -173,8 +183,7 @@ func handleBootstrapStopKeepalived(kubeconfigPath string, bootstrapStopKeepalive
 				"consecutiveErr":                consecutiveErr,
 				"bootstrapApiFailuresThreshold": bootstrapApiFailuresThreshold,
 			}).Info("handleBootstrapStopKeepalived: Num of failures exceeds threshold")
-			bootstrapStopKeepalived <- true
-			return
+			bootstrapStopKeepalived <- stopped
 		}
 		time.Sleep(1 * time.Second)
 	}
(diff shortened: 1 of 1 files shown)


pkg/monitor/dynkeepalived.go
```diff
@@ -166,14 +173,17 @@ func handleBootstrapStopKeepalived(kubeconfigPath string, bootstrapStopKeepalive
...
 			consecutiveErr = 0
 		}
 		if consecutiveErr > bootstrapApiFailuresThreshold {
 			log.WithFields(logrus.Fields{
 				"consecutiveErr":                consecutiveErr,
 				"bootstrapApiFailuresThreshold": bootstrapApiFailuresThreshold,
 			}).Info("handleBootstrapStopKeepalived: Num of failures exceeds threshold")
-			bootstrapStopKeepalived <- true
+			bootstrapStopKeepalived <- stopped
 			return
```
yboaron, line 187: We shouldn't exit this function after 'stop' message was sent to channel,

pkg/monitor/dynkeepalived.go
```diff
@@ -166,15 +173,17 @@ func handleBootstrapStopKeepalived(kubeconfigPath string, bootstrapStopKeepalive
...
 			consecutiveErr = 0
 		}
 		if consecutiveErr > bootstrapApiFailuresThreshold {
 			log.WithFields(logrus.Fields{
 				"consecutiveErr":                consecutiveErr,
 				"bootstrapApiFailuresThreshold": bootstrapApiFailuresThreshold,
 			}).Info("handleBootstrapStopKeepalived: Num of failures exceeds threshold")
-			bootstrapStopKeepalived <- true
-			return
+			bootstrapStopKeepalived <- stopped
```
yboaron, line 186: 'stop' message will be sent every second after API goes down, right?

pkg/monitor/dynkeepalived.go
```diff
@@ -306,23 +315,25 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
...
+				var cmdMsg []byte
+				if APIStateChanged == stopped {
+					cmdMsg = []byte("stop\n")
+				} else {
+					cmdMsg = []byte("reload\n")
+				}
+				_, err := conn.Write(cmdMsg)
 				if err == nil {
-					log.Info("Stop message successfully sent to Keepalived container control socket")
+					log.Info("Command message successfully sent to Keepalived container control socket: %s", string(cmdMsg[:]))
```
cybertron, line 329: This isn't formatting correctly. I think you need log.Infof to use the format string.

pkg/monitor/dynkeepalived.go
```diff
@@ -166,15 +173,17 @@ func handleBootstrapStopKeepalived(kubeconfigPath string, bootstrapStopKeepalive
 				"consecutiveErr": consecutiveErr,
 			}).Info("handleBootstrapStopKeepalived: detect failure on API")
 		} else {
+			if consecutiveErr > bootstrapApiFailuresThreshold { // Means it was stopped
```
cybertron, line 176: Ah, I believe this is why I added the separate start command, so I wouldn't have to check this. With a separate command I can just resend it safely and don't have to worry about a message getting lost because keepalived happens to restart or something.

This should be fine for the bootstrap though. It isn't a long-lived thing anyway.

cybertron (commented):
One minor thing about the log message, but otherwise this is working for me and should fix the problem.

pkg/monitor/dynkeepalived.go
yboaron: well, It doesn't matter because at the MCO side if there is no running Keepalived  process it just display a msg to log

celebdor:
/retitle Bug 1884420: bootstrap: API shows up, start it again

pkg/monitor/dynkeepalived.go
celebdor: right

yboaron:
I have one concern that is because we don't sync config change and stop bootstrap functions, so config change function can trigger  'reload' Keepalived in case of config change although it should be stopped.
But I assume after a second the handlestopbootstrap function will resend stop again .
so we should be good

celebdor:
> I have one concern that is because we don't sync config change and stop bootstrap functions, so config change function can trigger 'reload' Keepalived as a result of config change although Keepalived should be stopped.
> But I assume after a second the handlestopbootstrap function will resend stop .
> so we should be good
> 
> Since the masters don't know the bootstrap IP address we must verify that Keepalived on bootstrap isn't running otherwise two nodes will hold the VIP.

which two nodes?

PR #103, opened 2020-10-07
Bug 1885867 :  Flip to unicast only when MCO set to desired version in all nodes

yboaron:
In 4.5 to 4.6 upgrade we support automatic Keepalived mode flip to unicast,
the change to unicast should start only after MCO finished configuring the 4.6 desired state
in all the nodes.

With this PR the flip mode starts only after all the nodes configured with the new MCO configuration.

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -234,7 +234,9 @@ func IsUpgradeStillRunning(kubeconfigPath string) (error, bool) {
 	}
 
 	for _, node := range nodes.Items {
-		if node.Annotations["machineconfiguration.openshift.io/desiredConfig"] != node.Annotations["machineconfiguration.openshift.io/currentConfig"] {
+		if node.Annotations["machineconfiguration.openshift.io/desiredConfig"] != node.Annotations["machineconfiguration.openshift.io/currentConfig"] ||
+			node.Annotations["machineconfiguration.openshift.io/desiredConfig"] != nodes.Items[0].Annotations["machineconfiguration.openshift.io/desiredConfig"] {
+
 			return nil, true
 		}
 	}


pkg/config/node.go
```diff
@@ -234,7 +234,9 @@ func IsUpgradeStillRunning(kubeconfigPath string) (error, bool) {
 	}
 
 	for _, node := range nodes.Items {
-		if node.Annotations["machineconfiguration.openshift.io/desiredConfig"] != node.Annotations["machineconfiguration.openshift.io/currentConfig"] {
+		if node.Annotations["machineconfiguration.openshift.io/desiredConfig"] != node.Annotations["machineconfiguration.openshift.io/currentConfig"] ||
+			node.Annotations["machineconfiguration.openshift.io/desiredConfig"] != nodes.Items[0].Annotations["machineconfiguration.openshift.io/desiredConfig"] {
```
cybertron, line 238: So this is essentially checking that the current node matches the first node. Do we know that the first node will always be upgraded first or should we make this check a separate loop where we verify all of the nodes' desiredConfigs match (or I guess we could still use this loop with a bit different logic)?

yboaron: A. No, we don't know that the first node will be first.

B. This code supposes to check that : 
    * desiredConfig is equal to currentConfig for each node 
and 
   * all the nodes has the same desiredConfig 

So, If  desiredConfig != currentConfig for one of the nodes, we'll fail because of the first condition in the if statement. 

And the second condition in the if statement verifies that all the nodes has the same  desiredConfig.

I used the below logic to verify that all nodes has the same desiredConfig.
Assuming that we have an array of N (0 - N-1) numbers and we need to check if all values of the array are equal.

So, if for each i between 1 to N-1 Arr[i] == Arr[0] that means that all array members are equal (because whenever x = y and y = z, then also x = z )

Let me know if you still think I should change/update something in this statement
.

cybertron: Oh, right. This does check that they're all equal. Oops. :-)

cybertron (approved):
/lgtm

Sorry for the delay getting back to this.

PR #106, opened 2020-10-19
Add bz component information

celebdor:


OWNERS
```diff
@@ -8,3 +8,5 @@ approvers:
   - cybertron
   - bcrochet
   - yboaron
+component: Networking
+subcomponent: "runtime-cfg"
```
yboaron, line 12: Do we need the quotation in subcomponent ?

PR #108, opened 2020-11-04
Bug 1894539: Allow baremetal-runtimecfg to function when the node isn't attached to the VIP network

mandre:
There are a few problems preventing `baremetal-runtimecfg` to work when the node isn't attached to the VIP network.
- The `render` command complains when the node is not on a subnet on which the VIP is routable.
- The `node-ip` accepts running without providing a VIP, but still complains when we pass it a VIP and the node isn't attached to the subnet including the VIP.

This is a first step to support deploying additional machinesets on separate subnets.

### DIFF
--- a/cmd/runtimecfg/node-ip.go
+++ b/cmd/runtimecfg/node-ip.go
@@ -31,7 +31,7 @@ var nodeIPCmd = &cobra.Command{
 var nodeIPShowCmd = &cobra.Command{
 	Use:                   "show [Virtual IP...]",
 	DisableFlagsInUseLine: true,
-	Short:                 "Show a configured IP address that directly routes to the given Virtual IPs. If no Virtual IPs are provided it will pick an IP associated with the default route.",
+	Short:                 "Show a configured IP address that directly routes to the given Virtual IPs. If no Virtual IPs are provided or if the node isn't attached to the VIP subnet, it will pick an IP associated with the default route.",
 	Args:                  cobra.MinimumNArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		err := show(cmd, args)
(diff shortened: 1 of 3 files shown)


pkg/config/net.go
```diff
@@ -34,13 +39,36 @@ func GetInterfaceAndNonVIPAddr(vips []net.IP) (vipIface net.Interface, nonVipAdd
...
+	}
+	for _, iface := range ifaces {
+		addrs, err := iface.Addrs()
+		if err != nil {
+			return vipIface, nonVipAddr, err
+		}
+		for _, addr := range addrs {
+			switch n := addr.(type) {
+			case *net.IPNet:
+				if n.IP.String() == nodeAddrs[0].String() {
```
cybertron, line 65: I'm curious what would happen if AddressesDefault returned multiple matches, but I guess this only returns one address anyway so it probably doesn't matter.

cybertron (changes requested):
Overall lgtm. One issue with error handling inline.

pkg/config/net.go
mandre: AddressesDefault could return multiple matches if there are more than one interface configured on the subnet with the default routes. In this case we pick the first match, consistent with [how node-ip is implemented](https://github.com/openshift/baremetal-runtimecfg/blob/1cd2a95/cmd/runtimecfg/node-ip.go#L135). Do you think there's is a risk AddressesDefault returns results in a different order?

cmd/runtimecfg/node-ip.go
```diff
@@ -139,7 +139,8 @@ func getSuitableIP(retry bool, vips []net.IP) (chosen net.IP, err error) {
 
 		if len(vips) > 0 {
 			nodeAddrs, err = utils.AddressesRouting(vips, utils.ValidNodeAddress)
-		} else {
+		}
```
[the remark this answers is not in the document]
mandre: I added the missing err check in 384c503.

pkg/config/net.go
cybertron: I'm not sure. It's just that when I see something return an array and then the calling code only uses the first element it makes me wonder if there could be a bug hidden in there under some circumstances.

It's an existing problem with the function though. We're returning exactly one ip from calls that could return multiple in the existing code too, so I'm not worried about it in the context of this change. It was sort of a general observation.

PR #110, opened 2020-12-02
Update "runtimecfg node-ip" for dual-stack

danwinship:
In kube 1.20 you can pass two IPs to `kubelet --node-ip`, but we still need help from runtimecfg to find the _right_ IPs.

First commit is just a reorg, second commit adds unit tests without changing functionality, and then the third commit changes the utils functions to return dual-stack answers and changes runtimecfg to make use of it. (Previously the utils functions returned multiple IPs of the same family but the caller only ever used the first value, so this just changes it to return only that first value, plus a second value of the opposite family.)

(I still need to write the corresponding MCO change and test it all...)

### DIFF
--- a/cmd/runtimecfg/node-ip.go
+++ b/cmd/runtimecfg/node-ip.go
@@ -68,13 +68,13 @@ func show(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	chosenAddress, err := getSuitableIP(retry, vips)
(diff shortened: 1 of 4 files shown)


danwinship:
MCO change is https://github.com/openshift/machine-config-operator/pull/2277, and this is now tested.

cybertron (approved):
/lgtm

PR #111, opened 2020-12-09
Bug 1894539: Allow node-ip to function without attachment to VIP network

mandre:
Fall back to IP addresses associated with the default route in the when
the node is not on a subnet containing any of the VIPs.

Restore the feature implemented in
384c50336098baf83cb66c823cf142408d5ba8cf and broken with 83d05afd1b76b8f0078887f1001e722284235a3a.


danwinship:
oh, oops. I didn't realize that change had been made so I was thinking it was a mistake that it was doing fallback there
/lgtm

PR #112, opened 2020-12-09
Bug 1905233:  get the correct VRRP interface in case of IPV6 overlapping subnets

yboaron:
Since DHCPv6 assigns addresses with a /128, only the first 64 bits are considered when the VIP address is compared with the interface address, see [1].  
Comparing only the first 64 bits of the candidate address with the VIP address might lead for choosing the wrong interface in case of overlapping subnets , see [2] as an example .

To handle the overlapping IPV6 subnets case, we should also verify that the candidate address and VIP are L2 connected.

[1] https://github.com/openshift/baremetal-runtimecfg/blob/master/pkg/config/net.go#L39-#L41

[2]
2 interfaces:
Interface A subnet: 1001:db8::/120 , Interface B subnet: 1001:db8::f00/120
VIP address 1001:db8::64

### DIFF
--- a/cmd/runtimecfg/node-ip.go
+++ b/cmd/runtimecfg/node-ip.go
@@ -138,6 +138,8 @@ func set(cmd *cobra.Command, args []string) error {
 }
 
 func getSuitableIPs(retry bool, vips []net.IP) (chosen []net.IP, err error) {
+	// Enable debug logging in utils package
+	utils.SetDebugLogLevel()
 	for {
 		if len(vips) > 0 {
 			chosen, err = utils.AddressesRouting(vips, utils.ValidNodeAddress)
--- a/pkg/config/net.go
+++ b/pkg/config/net.go
@@ -33,13 +33,17 @@ func getInterfaceAndNonVIPAddr(vips []net.IP) (vipIface net.Interface, nonVipAdd
 				if _, ok := vipMap[n.IP.String()]; ok {
 					continue // This is a VIP, let's skip
 				}
-				// FIXME: Breaks if more then one interface on the host has the same prefix
-				// DHCPv6 assigns addresses with a /128 if encountered assume /64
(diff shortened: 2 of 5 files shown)


cybertron (changes requested):
Can we add a unit test case that covers this scenario? Now that we have a unit test framework for this code we might as well use it. :-)

yboaron:
> Can we add a unit test case that covers this scenario? Now that we have a unit test framework for this code we might as well use it. :-)

Hmm, I can't see how adding a unit test to utils/addresses.go may help here, the **utils/addresses.go** code handles correctly all CIDR values. the bug is in [1] code which wrongly hardcode subnet length to 64 before it checks if the candidate address contains the VIP address. so, since we still don't have unit test framework for config/net.go I don't see the added value of adding  
new test with subnet length 120 for example to utils/addresses.go.


[1] https://github.com/openshift/baremetal-runtimecfg/blob/master/pkg/config/net.go#L36-L43

bcrochet (commented):
/approve

celebdor:
I would really like a test that mocks the address retrieval showing addresses that match the bz and checks that we pick the right one.

yboaron:
> I would really like a test that mocks the address retrieval showing addresses that match the bz and checks that we pick the right one.

The utils/addresses.go unit tests were updated to use overlapping IPV6 subnets (as in BZ), do you think we should add something else to tests?

cybertron (commented):
The new unit tests are exactly what I had in mind for this, thanks!

I know Toni had some thoughts about alternate ways to handle the logging behavior, so I'm going to hold off on merging this until we've all had a chance to discuss that. Also, one minor thing inline about a comment that I think is outdated now.

cybertron:
/lgtm

Comment cleanup is done, logging concerns are addressed. Should be good to go.

PR #113, opened 2020-12-29
Move to quay.io dhcp image to avoid Dockerhub limits

YuviGold:


cybertron:
/lgtm

This will likely have to wait until after code freeze as it's not a bug fix per se, but it seems fine to me.

YuviGold:
There was an issue with the dhcpd volumed lease file, sometimes it couldn't be found by the dhcpd container.
Made the same changes as on https://github.com/openshift/assisted-installer-agent/pull/138

PR #117, opened 2021-01-31
Bug 1915080: add CloseIdleConnections for HTTP K8S API healthcheck

yboaron:
We noticed a linearly growing number of TCP connections in HAProxy in all clusters,
that are caused by periodic K8S API health check calls by the haproxy-monitor.
This PR verifies that unneeded HTTP connections used for checking API health being closed.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -80,6 +80,7 @@ func IsKubernetesHealthy(port uint16) (bool, error) {
 	if err != nil {
 		return false, err
 	}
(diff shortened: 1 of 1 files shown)


mandre (commented):
/lgtm
/retest
/bugzilla refresh

mandre:
/hold
Until the tests pass

mandre:
/lgtm

This seems much more reasonable now:
image

Compare this to https://github.com/openshift/machine-config-operator/pull/2350#issuecomment-763898294

PR #118, opened 2021-02-05
Add dnsmasqmonitor

cybertron:
This is for a simple monitor container that will simply watch the
cluster-hosted-net-services-operator configmap and update the api-int
record in dnsmasq appropriately. If it detects that the content of
the file has changed, it will overwrite the existing api-int.host
file and send the ClearCache dbus command to dnsmasq to reload its
configuration.


bcrochet (approved):
/lgtm

PR #124, opened 2021-03-02
Delete DNS-VIP and DNSVirtualRouterID          

yboaron:
DNS-VIP is not used anymore.

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -33,8 +33,6 @@ type Cluster struct {
 	APIVirtualRouterID     uint8
 	APIVIPRecordType       string
 	APIVIPEmptyType        string
-	DNSVIP                 string
-	DNSVirtualRouterID     uint8
 	IngressVIP             string
 	IngressVirtualRouterID uint8
 	IngressVIPRecordType   string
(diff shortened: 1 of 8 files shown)


cmd/runtimecfg/vr-ids.go
```diff
@@ -34,7 +34,7 @@ func runVRIds(cmd *cobra.Command, args []string) error {
 	clusterType := v.Type()
 	for i := 0; i < v.NumField(); i++ {
 		fieldName := clusterType.Field(i).Name
-		if strings.HasSuffix(fieldName, "RouterID") {
+		if strings.HasSuffix(fieldName, "RouterID") && fieldName != "DNSVirtualRouterID" {
```
cybertron, line 37: Shouldn't we just remove the DNS VIP fields from the config struct? They haven't been used in several releases now.

pkg/config/node.go
```diff
@@ -168,28 +166,21 @@ func (c *Cluster) PopulateVRIDs() error {
 		return fmt.Errorf("Cluster name can't be empty")
 	}
 	c.APIVirtualRouterID = utils.FletcherChecksum8(c.Name+"-api") + 1
-	c.DNSVirtualRouterID = utils.FletcherChecksum8(c.Name+"-dns") + 1
-	if c.DNSVirtualRouterID == c.APIVirtualRouterID {
-		c.DNSVirtualRouterID++
-	}
 	c.IngressVirtualRouterID = utils.FletcherChecksum8(c.Name+"-ingress") + 1
-	for c.IngressVirtualRouterID == c.DNSVirtualRouterID || c.IngressVirtualRouterID == c.APIVirtualRouterID {
+	for c.IngressVirtualRouterID == c.APIVirtualRouterID {
```
cybertron, line 170: Nit: This doesn't need to be a for loop anymore because there's only one other ID.

cybertron (commented):
Thanks! Will approve once it passes metal-ipi.

cybertron:
/retest

Looks like the cluster stood up fine. Some api test failed. :-/

cybertron:
/lgtm

metal-ipi is blocking, so if it doesn't pass this won't merge. Should be good to go though.

PR #125, opened 2021-03-02
Sort node-ip addresses based on default route priority

cybertron:
The existing implementation of node-ip lookup using default routes
may return different results if an address moves from a physical
interface to a bridge. If there are multiple default routes, this
may change the order in which we retrieve the addresses. Because we
only take the first one, this will cause us to select a different
address, which can break the node configuration.

To fix this, we need to sort the addresses based on the priority of
the associated default route. This will have no effect when there is
a single default route, but it will ensure we always return the
default route addresses in the same order no matter what order the
interfaces are in.

### DIFF
--- a/pkg/utils/addresses.go
+++ b/pkg/utils/addresses.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"net"
+	"sort"
 
 	"github.com/vishvananda/netlink"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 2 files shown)


cybertron:
/hold

Hmm, for some reason this is dropping addresses in the unit tests. Will need to look into that.

pkg/utils/addresses.go
```diff
@@ -213,8 +220,18 @@ func addressesDefaultInternal(af AddressFilter, getAddrs addressMapFunc, getRout
 			}
```
ori-amizur, line 220: Seems that code on line 218 should be updated

cybertron: Ah, yep. We're not populating matches in the loop anymore so that logic doesn't work.

cybertron:
/hold cancel

Unit tests are fixed.

osherdp:
@cybertron priority == metric?
if so, maybe we should add another level of determinism. I think that metric values can be the same in some cases, and sorting based on NIC names might be another thing we can do
WDYT?

osherdp:
Anyway, not an urgent issue I think. We can handle it afterwards

osherdp:
well,
/lgtm
:)

PR #129, opened 2021-03-17
Add node resolution to corednsmonitor

cybertron:
Now that node resolution is not needed for bootstrapping, we can
retrieve the list of nodes from the api in the coredns monitor
instead of having to maintain a coredns plugin. This adds a new
field to the cluster config that is populated with a list of the
nodes in the cluster, their addresses, and whether the address is
ipv6. This can be used in the coredns Corefile template to provide
node addresses statically.

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -26,6 +26,12 @@ const localhostKubeApiServerUrl string = "https://localhost:6443"
 
 var log = logrus.New()
 
+type NodeAddress struct {
+	Address string
+	Name    string
+	Ipv6    bool
+}
+
 type Cluster struct {
 	Name                   string
 	Domain                 string
@@ -39,6 +45,7 @@ type Cluster struct {
 	IngressVIPEmptyType    string
 	VIPNetmask             int
 	MasterAmount           int64
+	NodeAddresses          []NodeAddress
 }
 
 type Backend struct {
@@ -429,4 +436,52 @@ func GetClusterNameAndDomain(kubeconfigPath, clusterConfigPath string) (clusterN
 	}
 
 	return
+}
+
+func PopulateNodeAddresses(kubeconfigPath string, node *Node) {
+	// Get node list
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+	if err != nil {
+		log.Errorf("Failed to build client config: %s", err)
+		return
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		log.Errorf("Failed to create client: %s", err)
+		return
+	}
+	nodes, err := clientset.CoreV1().Nodes().List(metav1.ListOptions{})
+	if err != nil {
+		log.Errorf("Failed to get node list: %s", err)
+		return
+	}
+	for _, n := range nodes.Items {
+		name := ""
+		addr := net.IP{}
+		// Can't compare IP objects, need a separate flag
+		foundAddr := false
+		for _, a := range n.Status.Addresses {
+			if a.Type == v1.NodeHostName {
+				// We only want the shortname
+				name = strings.Split(a.Address, ".")[0]
+			} else if a.Type == v1.NodeInternalIP {
+				addr = net.ParseIP(a.Address)
+				foundAddr = true
+			}
+		}
+		if name == "" || !foundAddr {
+			log.Warningf("Could not handle node: %v", node)
+			continue
+		}
+		// TODO(bnemec): The ipv6 flag isn't currently used in the templates,
+		// but at some point it probably should be so we provide RFC-compliant
+		// ipv6 behavior.
+		ipv6 := true
+		check := addr.To4()
+		if check != nil {
+			ipv6 = false
+		}
+		node.Cluster.NodeAddresses = append(node.Cluster.NodeAddresses, NodeAddress{Address: addr.String(), Name: name, Ipv6: ipv6})
+	}
 }
--- a/pkg/monitor/corednsmonitor.go
+++ b/pkg/monitor/corednsmonitor.go
@@ -4,6 +4,7 @@ import (
 	"net"
 	"os"
 	"os/signal"
+	"sort"
 	"syscall"
 	"time"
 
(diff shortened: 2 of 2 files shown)


pkg/monitor/corednsmonitor.go
```diff
@@ -42,24 +43,34 @@ func CorednsWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath strin
 			if err != nil {
 				return err
 			}
```
yboaron, line 45: I checked deployment with both MCO and baremetal-runtimecfg PRs,  Corefile includes the nodes records.

I have two comments:

A. It could be useful to add log message with details about the cause for Corefile rendering.
B. Maybe it worth to add stabilization counter for changes before updating the Corefile, I checked coredns-monitor logs in my env [1] and it seems that Corefile was rendered 8 times .


[1] 

[core@master-1 ~]$ sudo crictl logs ac1436d7e367a
time="2021-03-18T11:17:35Z" level=error msg="Failed to build client config: invalid configuration: [unable to read client-cert /var/lib/kubelet/pki/kubelet-client-current.pem for default-auth due to open /var/lib/kubelet/pki/kubelet-client-current.pem: no such file or directory, unable to read client-key /var/lib/kubelet/pki/kubelet-client-current.pem for default-auth due to open /var/lib/kubelet/pki/kubelet-client-current.pem: no such file or directory]"
time="2021-03-18T11:17:35Z" level=error msg="Failed to build client config: invalid configuration: [unable to read client-cert /var/lib/kubelet/pki/kubelet-client-current.pem for default-auth due to open /var/lib/kubelet/pki/kubelet-client-current.pem: no such file or directory, unable to read client-key /var/lib/kubelet/pki/kubelet-client-current.pem for default-auth due to open /var/lib/kubelet/pki/kubelet-client-current.pem: no such file or directory]"
time="2021-03-18T11:18:05Z" level=info msg="Runtimecfg rendering template" path=/etc/coredns/Corefile
time="2021-03-18T11:18:35Z" level=info msg="Runtimecfg rendering template" path=/etc/coredns/Corefile
time="2021-03-18T11:20:43Z" level=error msg="Failed to get node list: etcdserver: request timed out"
time="2021-03-18T11:20:43Z" level=info msg="Runtimecfg rendering template" path=/etc/coredns/Corefile
time="2021-03-18T11:21:13Z" level=info msg="Runtimecfg rendering template" path=/etc/coredns/Corefile
time="2021-03-18T11:29:09Z" level=error msg="Failed to get node list: the server was unable to return a response in the time allotted, but may still be processing the request (get nodes)"
time="2021-03-18T11:29:09Z" level=info msg="Runtimecfg rendering template" path=/etc/coredns/Corefile
time="2021-03-18T11:29:39Z" level=info msg="Runtimecfg rendering template" path=/etc/coredns/Corefile
time="2021-03-18T11:35:39Z" level=info msg="Runtimecfg rendering template" path=/etc/coredns/Corefile
time="2021-03-18T11:36:09Z" level=info msg="Runtimecfg rendering template" path=/etc/coredns/Corefile
[core@master-1 ~]$

cybertron: > A. It could be useful to add log message with details about the cause for Corefile rendering.

Sure, I can do that.

> B. Maybe it worth to add stabilization counter for changes before updating the Corefile, I checked coredns-monitor logs in my env [1] and it seems that Corefile was rendered 8 times .

Ah, interesting. I think this is actually a bug. When it failed to reach the api server it re-rendered the config, which shouldn't happen.

The monitor only runs every 30 seconds, so I don't think it's likely to do much flapping. In a normal deployment I would expect up to 5 re-renders (assuming our default 5 node deployment) as each node becomes available. I would prefer to make sure the node list is as up to date as possible rather than try to limit the number of times we render the config. It's not that resource intensive and there's no outage (unlike keepalived and haproxy) when we do it.

I do need to fix the error handling so we aren't rendering empty node lists though.

celebdor: > Ah, interesting. I think this is actually a bug. When it failed to reach the api server it re-rendered the config, which shouldn't happen.

Does that mean that in such cases we'll keep the previously rendered nodes if they exist? Does this have any effect on power outages?

pkg/config/node.go
```diff
@@ -430,3 +437,45 @@ func GetClusterNameAndDomain(kubeconfigPath, clusterConfigPath string) (clusterN
...
+		return
+	}
+	for _, n := range nodes.Items {
+		name := ""
+		addr := net.IP{}
+		for _, a := range n.Status.Addresses {
+			if a.Type == v1.NodeHostName {
+				// We only want the shortname
+				name = strings.Split(a.Address, ".")[0]
+			}
```
yboaron, line 466: Wouldn't it be better to change it to 'if else' ?

pkg/config/node.go
```diff
@@ -430,3 +437,45 @@ func GetClusterNameAndDomain(kubeconfigPath, clusterConfigPath string) (clusterN
...
+		}
+		// TODO(bnemec): The ipv6 flag isn't currently used in the templates,
+		// but at some point it probably should be so we provide RFC-compliant
+		// ipv6 behavior.
+		ipv6 := true
+		check := addr.To4()
+		if check != nil {
+			ipv6 = false
+		}
```
yboaron, line 478: To be on the safe side, maybe we can append entry to node list only if : 
		name != = "" and 		addr != net.IP{} ?

pkg/monitor/corednsmonitor.go
yboaron: Yep, it looks better with the latest commit, the nodes entries were added to coredns Corefile and coredns file was rendered twice.

cybertron: > Does that mean that in such cases we'll keep the previously rendered nodes if they exist? Does this have any effect on power outages?

Yes, if we lose contact with the API then we'll leave the node list as it was the last time we were able to update. I would expect that to be better for recovery scenarios. Once the API comes back up we'll be able to update, but in the meantime all of the previous DNS records will still be available. This should better match what happens on other platforms where the DNS records come from the cloud DNS instead of in-cluster.

pkg/config/node.go
cybertron: Sure, I can do that.

pkg/config/node.go
cybertron: I feel like it would be an error if we got a node record that looked like that, but it shouldn't hurt to add a check.

pkg/config/node.go
yboaron: Thanks

PR #130, opened 2021-03-19
Bug 1939740: sort AddressesDefault by ifindex and IPv4/IPv6 preference

danwinship:
Two tweaks to `node-ip` behavior in the UPI case:
- If there are two interfaces with equal-priority default routes, use the IPs from the one with the lower ifindex. AFAIK this isn't actually affecting anyone yet, but if someone _did_ do this, then with the current code we'd randomly flip between the interfaces/addresses based on the order golang happened to iterate the map in. This makes us not do that.
- This adds `--prefer-ipv6` so you can clarify whether you want an IPv4 or IPv6 address for single-stack kubernetes on dual-stack nodes (which is what the linked bz is)

### DIFF
--- a/cmd/runtimecfg/node-ip.go
+++ b/cmd/runtimecfg/node-ip.go
@@ -20,7 +20,7 @@ const (
 	crioSvcOverridePath    = "/etc/systemd/system/crio.service.d/20-nodenet.conf"
 )
 
-var retry bool
+var retry, preferIPv6 bool
 var nodeIPCmd = &cobra.Command{
 	Use:                   "node-ip",
 	DisableFlagsInUseLine: true,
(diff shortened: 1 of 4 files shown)


celebdor:
/hold
Until there's some manual testing

danwinship:
```
[root@worker-1 core]# /usr/bin/podman run --rm   --authfile /var/lib/kubelet/config.json   --net=host   --volume /etc/systemd/system:/etc/systemd/system:z   registry.build01.ci.openshift.org/ci-ln-pyzst3b/stable@sha256:8fa36383c573611e966c8d88345e9a05d431147e8adaab4738170cc26270d929 node-ip show 2>/dev/null | tail -1
192.168.111.24
[root@worker-1 core]# /usr/bin/podman run --rm   --authfile /var/lib/kubelet/config.json   --net=host   --volume /etc/systemd/system:/etc/systemd/system:z   registry.build01.ci.openshift.org/ci-ln-pyzst3b/stable@sha256:8fa36383c573611e966c8d88345e9a05d431147e8adaab4738170cc26270d929 node-ip show --prefer-ipv6 2>/dev/null | tail -1
fd2e:6f44:5dd8:c956::18
```
/hold cancel

PR #133, opened 2021-03-24
Bug 1942506: 4.6 EUS node-ip handling extravaganza

danwinship:
This started because I need to backport #130 to 4.6, but it didn't backport cleanly at all because there have been a bunch of node-ip changes since then...

Given that 4.6 is our extended support release, it is probably good to get all the bug-fix-y bits backported. We don't actually need the dual-stack support in 4.6 (and I don't plan to backport the corresponding MCO part of that) but including it makes it easier to backport the rest of it... (We do expect to have at least one customer running single-stack IPv6 for a while in 4.6, hence the need for the #130 backport.)

/cc @cybertron @celebdor


cybertron:
/lgtm

Agree that backporting the dual stack support to allow for clean backports of the bugfixes is the best approach. Trying to untangle all those changes would be very error-prone.

PR #134, opened 2021-03-25
Remove etcd name logic

cybertron:
We no longer use etcd hostnames or DNS records, so we don't need
to have this logic in runtimecfg.


cybertron:
/retest

Looks like a simple timeout

PR #135, opened 2021-04-23
Bug 1949859: Add Ironic healthcheck to bootstrap keepalived checks

cybertron:
Ironic has started using the API VIP for access, and as a result we
need to keep the VIP available when Ironic is. The apiserver can go
down periodically during bootstrapping, so that isn't a reliable way
to keep the VIP alive on its own.

This change adds another check when the apiserver check fails. It
will ping Ironic and if Ironic is still up then it won't count the
check as failed and will keep the VIP active.

Note that previously Ironic was never stopped on the boostrap node.
Doing that will be a pre-requisite to this functionality as we do
need to allow keepalived to stop once bootstrapping is complete so
the VIP moves to the masters. Since bootstrapping can't complete
until after Ironic has provisioned the masters, if we stop Ironic
after provisioning then this check won't prevent VIP migration when
bootstrap is complete.

### DIFF
--- a/pkg/monitor/dynkeepalived.go
+++ b/pkg/monitor/dynkeepalived.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"net/http"
 	"os"
 	"os/signal"
 	"strings"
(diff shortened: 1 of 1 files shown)


cybertron:
/hold

This won't work without changes to the Ironic lifecycle on the boostrap node.

stbenjam:
/retitle Bug 1949859: Add Ironic healthcheck to bootstrap keepalived checks

cybertron:
Once https://github.com/openshift/installer/pull/4900 merges this should pass ci.

hs0210:
Manually merging [#135](https://github.com/openshift/baremetal-runtimecfg/pull/135) and [#4900](https://github.com/openshift/installer/pull/4900), I did several IPI tests, this problem did not reappear.

zhouhao3:
https://github.com/openshift/installer/pull/4900  has been merged. It seems that this PR needs a retest.

rhjanders:
Thank you @zhouhao3, just requested retest.

rhjanders:
/recheck again just to see if we get the same failures again

PR #136, opened 2021-05-09
Keepalived - don't configure empty peers list on bootstrap

yboaron:
We should run the following steps for Keepalived on bootstrap:
1. Configure empty peers list ( bootstrap should hold the VIP first)
2. After masters registered to cluster , add masters to peers list 
3. Stop Keepalived when local kube-apiserver stop running 

We shouldn't allow configuring empty peers list after step 2 (it will cause 2 separate VRRP  domains --> two nodes may hold the VIP simultaneously )

### DIFF
--- a/pkg/monitor/dynkeepalived.go
+++ b/pkg/monitor/dynkeepalived.go
@@ -38,6 +38,8 @@ const (
 	started APIState = iota
 )
 
+var BootstrapPeersListConfiguredOnce bool = false
+
 func getActualMode(cfgPath string) (error, bool) {
 	enableUnicast := false
 	_, err := os.Stat(cfgPath)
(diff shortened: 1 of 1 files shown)


cybertron (commented):
Okay, as I was retitling this I had second thoughts. My expectation of what was happening is that a master was starting keepalived with no peers and thus taking the API VIP, which left it unable to update the peer list when the nodes registered and so it was stuck permanently in multicast mode. However, I realized that we already don't allow masters to come up without unicast peers and I'm having trouble understanding why it matters on the bootstrap. Even if the bootstrap temporarily came up with no peers, it should quickly populate that list and coordinate with the other nodes. At worst you might have the VIP in two places for a few seconds until keepalived sorts itself out. That doesn't fit the behavior I've observed when I ran into this locally though.

I'm going to try to wrap my head around what's going on here for a little longer, but if I can't do that I'll go ahead and merge it since ci suggests it is helping.

/retitle Bug 1957708: Keepalived - don't configure empty peers list on bootstrap
/lgtm cancel

zhouhao3:
I also encountered the same problem when using IPI to deploy the OCP environment on baremetal. During the deployment process, both bootstrap and master0 had apivip, which caused the deployment to fail.
I merged your patch and tested it again. The error did not reappear in the first three times, and the problem reappeared in the fourth and fifth tests.

yboaron:
@zhouhao3 
Maybe you can try https://github.com/openshift/machine-config-operator/pull/2576

cybertron:
/retitle Keepalived - don't configure empty peers list on bootstrap

This was fixed by #137, so unlinking this from the bz so it can proceed.

PR #141, opened 2021-06-02
Bug 1886572: Calculate keepalived priority for ingress

yboaron:
Ingress VIP should be set only on a node that runs an instance of the default ingress controller pod.
    In current code, in case extra ingress-controllers are created the ingress VIP might be wrongly set on a node that doesn't run
    an instance of the default ingress controller.
    
    This PR calculates the priority for keepalived ingress VIP depending on the presence of the router pod in the node by monitoring the content of router-internal-default endpoints resource.

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -22,7 +22,10 @@ import (
 	"github.com/openshift/installer/pkg/types"
 )
 
-const localhostKubeApiServerUrl string = "https://localhost:6443"
+const (
+	localhostKubeApiServerUrl    = "https://localhost:6443"
+	defaultIngressControllerName = "router-internal-default"
+)
 
 var log = logrus.New()
 
@@ -63,7 +66,8 @@ type ApiLBConfig struct {
 }
 
 type IngressConfig struct {
-	Peers []string
+	Peers    []string
+	Priority int
 }
 
 type Node struct {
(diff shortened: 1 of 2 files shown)


yboaron:
/retitle Bug 1886572: Calculate keepalived priority for ingress

cybertron:
Is there any chance we can do this in a check script instead of the monitor? Or do we not have the ability to run oc commands from inside the keepalived container? It would just be nice to keep all of the priority bits in the same place so we're not having to look at both the keepalived and monitor logs when trying to figure out what happened with the priority if/when a VIP ends up somewhere it shouldn't.

Although maybe on reload keepalived would log the new priority anyway? Mostly I want to avoid making keepalived harder to debug than it already is.

yboaron:
> Is there any chance we can do this in a check script instead of the monitor? Or do we not have the ability to run oc commands from inside the keepalived container? It would just be nice to keep all of the priority bits in the same place so we're not having to look at both the keepalived and monitor logs when trying to figure out what happened with the priority if/when a VIP ends up somewhere it shouldn't.
> 
> Although maybe on reload keepalived would log the new priority anyway? Mostly I want to avoid making keepalived harder to debug than it already is.

Agree that having check_script for this purpose is ideal, we need to be able :
1. As you mentioned, to run oc command from the keepalived container 
2. Retrieve node's IP address

I'll try to mount host's oc binary and check that

pkg/config/node.go
```diff
@@ -218,6 +222,30 @@ func IsUpgradeStillRunning(kubeconfigPath string) (error, bool) {
 	return nil, false
 }
 
+func GetIngressPriority(kubeconfigPath string,nonVirtualIP string) (int){
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+	if err != nil {
+		return 40
```
cybertron, line 228: I don't think we want to default to 40 for the error case. If we can't determine whether the node has the default ingress we should give it the lower priority so we don't take the VIP from a node that is known to have the right ingress.

This applies to all of the other error cases below too.

Also, can we log the error? Otherwise all we know is that we got a priority of 20, but that could mean we didn't have the ingress or it could mean something went wrong here.

cybertron (changes requested):
After some discussion elsewhere, we decided to proceed with this solution since it's the one we have implemented and code freeze is coming up.

However, I have concerns about the error handling and we need to fix the fmt errors before it can go in.

pkg/config/node.go
yboaron: I set the priority to 40 in case of error because I didn't want to change current behavior (where priority set to 40 by default) though setting the priority to 20 in error case makes sense.

I'll change it to 20.

yboaron:
retest

cybertron:
/lgtm

The metal-ipi job failure appears unrelated and the ipv6 job passed so this should be fine.

PR #145, opened 2021-06-22
Bug 1974350:  HAProxy-monitor: send reload only if cfg file changed

yboaron:


stbenjam:
/retitle Bug 1974350:  HAProxy-monitor: send reload only if cfg file changed

PR #150, opened 2021-08-23
Bug 1925542: Add duplicate MAC check to DHCP VIP lease

cybertron:
We want to ensure that we never attempt to reuse an actual local
MAC address for the DHCP VIP lease. This adds a check to the
handleLeasing function that will error if a physical interface on
the system is found with the same MAC specified in the VIP config
file.

### DIFF
--- a/pkg/monitor/dynkeepalived.go
+++ b/pkg/monitor/dynkeepalived.go
@@ -236,6 +236,19 @@ func handleLeasing(cfgPath string, apiVip, ingressVip net.IP) error {
 		return fmt.Errorf("Mismatched ip for ingress. Expected: %s Actual: %s", ingressVip.String(), vips.IngressVip.IpAddress)
 	}
 
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


cybertron:
/hold

I'm not sure this is ready, but I'm having trouble running the unit tests locally and want to see if the same thing happens here.

cybertron:
I'm going to leave the hold because we need to get this tested with a DHCP VIP deployment. I _think_ it should work though.

dominikholler:
Thanks for the PR!
According to my understanding, which might be very incomplete, the relevant interface is created in https://github.com/openshift/assisted-installer-agent/blob/efe8ea006c178db1d8a6164262825219c9eb7fa4/src/dhcp_lease_allocate/lease.go#L29 using the same mac on multiple nodes.

cybertron:
I'm going to close this for now. It's probably not a bad idea, but it doesn't sound like it would solve the reported problem and it adds a bunch more logic that could have unexpected consequences. We can always reopen it if there ends up being a lot of issues with duplicated MACs due to configuration errors.

PR #154, opened 2021-09-19
Bug 1998951: retrieve only single type  addresses for Keepalived ingress

yboaron:
In case of dual-stack deployment, configure unicast_peer of ingress
VRRP instance with single type (v4/v6) IPs

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -218,7 +218,7 @@ func IsUpgradeStillRunning(kubeconfigPath string) (error, bool) {
 	return nil, false
 }
 
-func GetIngressConfig(kubeconfigPath string) (ingressConfig IngressConfig, err error) {
+func GetIngressConfig(kubeconfigPath string, filterIpType string) (ingressConfig IngressConfig, err error) {
 	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
 	if err != nil {
 		return ingressConfig, err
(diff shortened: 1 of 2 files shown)


pkg/config/node.go
```diff
@@ -235,11 +235,18 @@ func GetIngressConfig(kubeconfigPath string) (ingressConfig IngressConfig, err e
 	}
 
 	for _, node := range nodes.Items {
+		nodeIp := ""
 		for _, address := range node.Status.Addresses {
 			if address.Type == v1.NodeInternalIP {
-				ingressConfig.Peers = append(ingressConfig.Peers, address.Address)
+				nodeIp = address.Address
```
cybertron, line 241: I think we also need to check that we pick the IP version matching the VIP. I'm not sure selecting an IPv6 internal IP for an IPv4 VIP will work.

Even if it does, it will still trigger the warning message in the keepalived logs which means we'll keep getting bugs opened about this. :-)

cybertron (changes requested):

pkg/config/node.go
yboaron: Well, I also thought picking the IP matching the VIP but it seems that picking the first IP for each node works for LB details [1] for dual-stack case.

To be on the safe side, I changed the code to pick IP according to VIP


[1]
https://github.com/openshift/baremetal-runtimecfg/blob/master/pkg/config/node.go#L369-#L382

cybertron (approved):
/retest

This version lgtm. Would like to see the dualstack job pass though.

cybertron:
/retest

Failure does not appear related to me.

PR #155, opened 2021-10-11
Bug 2012780: Avoid dynamically allocated port range for haproxy

cybertron:
The haproxy stats and health endpoints were configured with high
port numbers (50000 and 50936) that conflict with the range used for
dynamic allocation of ports:

$ cat /proc/sys/net/ipv4/ip_local_port_range
32768   60999

This can occasionally cause haproxy to fail if the kernel happens to
pick one of those ports for a client connection.

To avoid this, we should use ports outside the range shown above.
This patch changes the stats endpoint to 30000, while the health
endpoint is controlled by the MCO config and will be fixed in a
separate commit there.


cybertron:
Bah. This version should catch all of the places that value is specified.

PR #157, opened 2021-11-09
Bug 2021544: Ignore VIPs in node-ip show

cybertron:
Currently the logic in node-ip show allows a VIP to be selected as
the node IP. This is incorrect, although it doesn't generally cause
a problem today because it appears the VIP is ordered after the
correct node IP in the list. However, we shouldn't rely on that
ordering to ensure we pick the correct address.

Note that this was discovered due to a misconfiguration in a cluster
where the VIP was placed on a node that had an address in a different
subnet. In this case, the node IP didn't match so the VIP was
subsequently checked and found to match (because of course it does).
This caused the VIP to be selected as the node IP, which is also an
error. While this scenario is invalid, it does show that it is
possible for the VIP to be selected as a node IP and we should not
allow that.

This patch checks each address in addressesRoutingInternal
against the VIP list and skips processing if it matches a VIP. This
should ensure we never select a VIP as a node IP.

### DIFF
--- a/pkg/utils/addresses.go
+++ b/pkg/utils/addresses.go
@@ -135,6 +135,16 @@ func addressesRoutingInternal(vips []net.IP, af AddressFilter, getAddrs addressM
 	for link, addresses := range addrMap {
 	addrLoop:
 		for _, address := range addresses {
+			isVip := false
+			for _, vip := range vips {
+				if address.IP.String() == vip.String() {
+					log.Debugf("Address %s is VIP %s. Skipping.", address, vip)
+					isVip = true
+				}
+			}
+			if isVip {
+				continue
+			}
 			maskPrefix, maskBits := address.Mask.Size()
 			if net.IPv6len == len(address.IP) && maskPrefix == maskBits {
 				if routeMap == nil {


yboaron:
Hmmm, what do you expect to happen with such deployment with this patch? should the deployment fail? what error? 
If it's an invalid configuration, I think we better block it in the installation phase.

cybertron:
> Hmmm, what do you expect to happen with such deployment with this patch? should the deployment fail? what error? If it's an invalid configuration, I think we better block it in the installation phase.

We do have a validation in the installer to make sure the VIP is part of machineCIDR. I think what happened in this case is that a machineset was added after installation with incorrect networking. I'm not entirely sure what we can do about that though. It's actually a valid use case for remote workers.

To be clear, this patch isn't so much about fixing anything as it is making errors easier to debug. The cluster in question had a misconfiguration, but runtimecfg picking the VIP as the node IP made the problem more confusing to debug because we spent a bunch of time figuring out why. I think what should have happened is that the VIP would be inaccessible because it was on a node on a different subnet. While that's still not ideal, at least we'd be looking at the right thing (the network configuration) when debugging.

cybertron:
/retest
Why does metal-ipi hate me? :-P

yboaron:
OK, if this makes debugging easier in such cases, let's merge it.
/lgtm

PR #158, opened 2021-11-16
Bug 2022050: Add delay after sending bootstrap stop and start messages

cybertron:
In some environments we seem to be hitting an edge case where more
than one message backs up on the bootstrapStopKeepalived queue,
which means it is possible for a stop and reload message to be sent
back to back. It appears that the stop function on the keepalived
side takes longer, and as a result reload completes before stop
and it sends a SIGHUP to a process that's about to be killed instead
of restarting keepalived like we want.

This patch adds a 5 second delay after a stop or reload message is
sent so keepalived should have plenty of time to process it before
another message comes in.

### DIFF
--- a/pkg/monitor/dynkeepalived.go
+++ b/pkg/monitor/dynkeepalived.go
@@ -316,6 +316,9 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
 				}).Error("Failed to write command to Keepalived container control socket")
 				time.Sleep(1 * time.Second)
 			}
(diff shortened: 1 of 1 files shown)


cybertron:
/bugzilla refresh

Trying again since the bot seems to have ignored my last attempt...

yboaron:
Well, usually solving race issues using delay is not the best idea, since it is only relevant to the bootstrap phase, I guess we can continue with the delay.
We have enough time until release to make sure this change does not affect other flows.
/lgtm

yboaron (approved):

PR #159, opened 2021-11-22
Fix docker_test run

creydr:
The `docker_test` target was failing because of some outdated image references. This PR fixes it so `make docker_test` can run again


cybertron:
/lgtm
/approve

We should talk to the assisted installer team and see if they're still using this. If so, there should be ci for it. The fact that it's broken (and probably has been for a while) makes me think nobody's using it though.

PR #162, opened 2021-11-25
Get cluster name and base domain from api server

creydr:
Currently we extract the cluster name and cluster domain from the kubeconfig file (on the cluster nodes). This is done by splitting the server url by `.`: https://github.com/openshift/baremetal-runtimecfg/blob/92ec13e1d0daf6eb644cfe25271f341e1fce78ff/pkg/config/node.go#L111-L127

As [BZ 1971709](https://bugzilla.redhat.com/show_bug.cgi?id=1971709) shows this can lead to problems, if the cluster name itself contains a dot. This PR addresses this and gets the cluster name and cluster domain from the `cluster-config-v1` config map, which contains the install-config and the cluster name and cluster domain.

### DIFF
--- a/cmd/monitor/monitor.go
+++ b/cmd/monitor/monitor.go
@@ -21,7 +21,7 @@ func main() {
 				cmd.Help()
 				return nil
 			}
-			clusterName, clusterDomain, err := config.GetKubeconfigClusterNameAndDomain(args[0])
+			clusterName, clusterDomain, err := config.GetClusterNameAndDomainFromAPIServer(args[0])
 			if err != nil {
 				return err
 			}
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
-	"net/url"
 	"os"
 	"sort"
 	"strings"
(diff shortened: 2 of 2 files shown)


creydr:
@cybertron & @yboaron: how do you usually test and validate changes in baremetal-runtimecfg?

cybertron:
At the moment most of the testing on this repo is e2e only. We have some unit tests, but I don't see much that covers this part of the code.

However, it looks like both e2e jobs are failing on problems rendering the coredns and keepalived config. It's possible that we don't have what we need to talk to the API at that point in the deployment process. If you want to try the changes out locally to see if it reproduces and poke around the environment, you can follow https://github.com/openshift-metal3/dev-scripts#testing-custom-container-images to deploy the patch locally.

creydr:
> However, it looks like both e2e jobs are failing on problems rendering the coredns and keepalived config. It's possible that we don't have what we need to talk to the API at that point in the deployment process. If you want to try the changes out locally to see if it reproduces and poke around the environment, you can follow https://github.com/openshift-metal3/dev-scripts#testing-custom-container-images to deploy the patch locally.

@cybertron thanks a lot for pointing me to the documenation with custom images. This helped me a lot to debug further. It seems we have a "running" cluster at that point. Anyhow the provided kubeconfig `/etc/kubernetes/kubeconfig` seems not to have a user configured with proper permissions:

```
Error: could not get cluster-config-v1 config map: configmaps "cluster-config-v1" is forbidden: User "system:serviceaccount:openshift-machine-config-operator:node-bootstrapper" cannot get resource "configmaps" in API group "" in the namespace "kube-system"
```

When I replaced this kubeconfig with one provided from the bootstrap node from `/var/opt/openshift/auth/kubeconfig`, I didn't get the permission errors. After I changed the static pod manifests from coredns, haproxy & keepalived to use this config (some were referencing `/var/lib/kubelet/kubeconfig`) on all master nodes, the cluster came up. 

Do you know, how we can get a kubeconfig with a user with proper permissions deployed on the master nodes?

cybertron:
Unfortunately I don't know. We didn't have anything to do with creating those kubeconfigs, we just consume them.

Did you try switching keepalived to the one in /var/lib though? They're not the same, and in my experience the one in /var/lib/kubelet has more permissions. See https://github.com/openshift/machine-config-operator/pull/2465#discussion_r605732694. We had to use that to get the node list and the same might work for this config map.

PR #164, opened 2021-12-06
Bug 1995021: upgrade k8s.io/client-go

yboaron:
We noticed that after VIP switchover coredns-monitor is stuck on get nodes code [1] for ~20 minutes, upgrade k8s.io/client-go fixed this issue. 

[1] https://github.com/openshift/baremetal-runtimecfg/blob/master/pkg/config/node.go#L455


cybertron:
/lgtm

Paging the ci bot...

yboaron:
This PR solves one of the regression tests that constantly fail in QE, checking if auto cherry-pick is possible here
/cherry-pick release-4.9

PR #173, opened 2022-03-24
[WIP - need test] node: update IsUpgradeStillRunning() logic

dougsland:
Currently, IsUpgradeStillRunning() ignore the fact master and
workers nodes contain different configs type.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2053309


dougsland:
I will submit a new patch, setting as hold.
/hold

pkg/config/node.go
```diff
@@ -191,32 +196,105 @@ func GetVRRPConfig(apiVip, ingressVip net.IP) (vipIface net.Interface, nonVipAdd
...
+
+// IsTheSameConfig will compare the config annotations
+//
+// Args:
+//      - The node list based on v1.NodeList
+//
+// Returns:
+//      - error or true (same config) or false
+//
+func IsTheSameConfig(nodes []v1.Node) (bool, error) {
```
cybertron, line 255: We don't need to return error here since it's never anything but nil.

pkg/config/node.go
```diff
@@ -191,32 +196,105 @@ func GetVRRPConfig(apiVip, ingressVip net.IP) (vipIface net.Interface, nonVipAdd
...
+
+	desiredConfig := "machineconfiguration.openshift.io/desiredConfig"
+	currentConfig := "machineconfiguration.openshift.io/currentConfig"
+
+	for _, node := range nodes {
+		if node.Annotations[desiredConfig] != node.Annotations[currentConfig] ||
+			node.Annotations[desiredConfig] != nodes[0].Annotations[desiredConfig] {
+			return true, nil
```
cybertron, line 263: I think this is backward. It's going to return true if the configs differ.

pkg/config/node.go
```diff
@@ -191,32 +196,105 @@ func GetVRRPConfig(apiVip, ingressVip net.IP) (vipIface net.Interface, nonVipAdd
...
+
+	// Go to all node types identified in GetNodes()
+	for instance := range nodes {
+		nodesConfigs, err := IsTheSameConfig(nodes[instance])
+		if err != nil {
+			return false, err
+		}
 
-			return nil, true
+		if nodesConfigs == true {
```
cybertron, line 292: Nit: nodesConfigs is already boolean so you don't need to compare it. Just "if nodesConfigs {" will do.

Also, related to my comment above I think this should be inverted. If all nodes have the same config, then the upgrade is not running anymore. It works right now because it's backward in both places, but it's confusing because of the name of the function.

pkg/config/node.go
```diff
@@ -191,32 +196,105 @@ func GetVRRPConfig(apiVip, ingressVip net.IP) (vipIface net.Interface, nonVipAdd
...
+		return nil, err
 	}
 
 	for _, node := range nodes.Items {
-		if node.Annotations["machineconfiguration.openshift.io/desiredConfig"] != node.Annotations["machineconfiguration.openshift.io/currentConfig"] ||
-			node.Annotations["machineconfiguration.openshift.io/desiredConfig"] != nodes.Items[0].Annotations["machineconfiguration.openshift.io/desiredConfig"] {
+		// Get all labels for *each node* which contain node-role.kubernetes.io/
+		for k, _ := range node.Labels {
+			switch {
+			case strings.HasPrefix(k, labelNodeRolePrefix):
```
cybertron, line 237: Is there a reason this is a switch instead of simple if?

pkg/config/node.go
```diff
@@ -23,6 +23,11 @@ import (
 	"github.com/openshift/installer/pkg/types"
 )
 
+const (
```
cybertron, line 26: Nit: This section could be combined with the const below.

cybertron (changes requested):
This did work in my local test, but I think there's some backward logic based on the name and docstring for the new function. It works because the consuming code is also backward, but I think we should fix it to avoid confusion.

A few other nits too, but nothing that affects the functionality of the change.

pkg/config/node.go
dougsland: We never know which labels we will look for sure in the future. Another point, I was expecting this label as well:  "kubernetes.io/role" so a switch make code simple and cleaner in case of multiple labels to look.

pkg/config/node.go
dougsland: done

pkg/config/node.go
dougsland: done

pkg/config/node.go
dougsland: done

pkg/config/node.go
dougsland: done

pkg/config/node.go
```diff
@@ -191,32 +195,104 @@ func GetVRRPConfig(apiVip, ingressVip net.IP) (vipIface net.Interface, nonVipAdd
...
+//	etc.
+//
+// Args:
+//      - kubeconfigPath as string
+//
+// Returns:
+//      - map[string][]v1.Node or error
+//
+func GetNodes(kubeconfigPath string) (map[string][]v1.Node, error) {
+
```
creydr, line 214: nit: remove empty line at beginning

pkg/config/node.go
```diff
@@ -191,32 +195,104 @@ func GetVRRPConfig(apiVip, ingressVip net.IP) (vipIface net.Interface, nonVipAdd
...
 	if err != nil {
-		return err, true
+		return nil, err
 	}
 
 	for _, node := range nodes.Items {
-		if node.Annotations["machineconfiguration.openshift.io/desiredConfig"] != node.Annotations["machineconfiguration.openshift.io/currentConfig"] ||
-			node.Annotations["machineconfiguration.openshift.io/desiredConfig"] != nodes.Items[0].Annotations["machineconfiguration.openshift.io/desiredConfig"] {
+		// Get all labels for *each node* which contain node-role.kubernetes.io/
+		for k, _ := range node.Labels {
```
creydr, line 234: nit: rename `k` to something like `label`. In addition `, _` should not be needed in this case.

pkg/config/node.go
```diff
@@ -191,32 +195,104 @@ func GetVRRPConfig(apiVip, ingressVip net.IP) (vipIface net.Interface, nonVipAdd
...
+// IsTheSameConfig will compare the config annotations
+//
+// Args:
+//      - The node list based on v1.NodeList
+//
+// Returns:
+//      - true (same config) or false
+//
+func IsTheSameConfig(nodes []v1.Node) bool {
+
```
creydr, line 255: nit: remove empty line at beginning

pkg/config/node.go
```diff
@@ -191,32 +195,104 @@ func GetVRRPConfig(apiVip, ingressVip net.IP) (vipIface net.Interface, nonVipAdd
...
+// comparing the node's state.
+//
+// Args:
+//      - kubeconfigPath as string
+//
+// Returns:
+//      - true (same config), false or error
+//
+func IsUpgradeStillRunning(kubeconfigPath string) (bool, error) {
+
```
creydr, line 278: nit: remove empty line at beginning

pkg/config/node.go
```diff
@@ -191,32 +195,104 @@ func GetVRRPConfig(apiVip, ingressVip net.IP) (vipIface net.Interface, nonVipAdd
...
+	nodes, err := GetNodes(kubeconfigPath)
+	if err != nil {
+		return false, err
+	}
+
+	// Go to all node types identified in GetNodes()
+	for instance := range nodes {
```
creydr, line 285: not sure if something like `nodeRole` would be more speaking instead of `instance` here :man_shrugging:

cybertron:
/retest
/lgtm

I tested this locally and it still works. My comments have been addressed, so we just need to win the ci lottery to get it landed. ;-)

dougsland:
let me address the comment from Christoph
/hold

pkg/config/node.go
dougsland: make sense, I was looking for a better name.

creydr (approved):
/lgtm

PR #174, opened 2022-03-29
Dockerfile: migrate repo to CentOS 8 Stream

dougsland:
Avoid errors related to appstream repo.


dougsland:
@cybertron worked for me.

cybertron:
/lgtm
/approve

Thanks! This also worked for me locally. It's not tested by ci so we don't need to wait for any optional jobs to pass.

PR #175, opened 2022-03-29
Bug 2069740: Avoid kubernetes node port range

cybertron:
Per [0], the range 30000-32767 is used for kubernetes node ports.
We should not have haproxy listening in that range to avoid possible
conflicts.

0: https://github.com/openshift/openshift-docs/blob/main/modules/installation-network-user-infra.adoc


cybertron:
/hold

It's been suggested that we make this 20000 more than the actual service port (9445 in this case).

creydr (approved):
/lgtm

PR #177, opened 2022-04-12
Update OWNERS to reflect current team

cybertron:
Toni and Yossi have moved to different teams so they aren't reviewing
in this repo anymore. We also have Christoph and Douglas working on
this now so we should have them on the list.


dougsland (commented):
/lgtm

PR #181, opened 2022-06-13
Bug 2096226: Check chosen node-ip can be used

derekhiggins:
In IPv6, if the chosen IP address is still tentative
then it can't be bound to, resulting in crio crashing.
Check that the IP address can be used and if not fail
or retry as appropriate.

### DIFF
--- a/cmd/runtimecfg/node-ip.go
+++ b/cmd/runtimecfg/node-ip.go
@@ -145,6 +145,20 @@ func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool) (chosen []net.IP
 		if len(vips) > 0 {
 			chosen, err = utils.AddressesRouting(vips, utils.ValidNodeAddress)
 			if len(chosen) > 0 || err != nil {
+
+				// If using IPv6, verify that the choosen address isn't tentative
+				// i.e. we can actually bind to it
+				if len(chosen) > 0 && net.IPv6len == len(chosen[0]) {
+					_, err := net.Listen("tcp", "["+chosen[0].String()+"]:")
+					if err != nil {
+						log.Errorf("Chosen node IP is not usable")
+						if !retry {
+							return nil, fmt.Errorf("Failed to find node IP")
+						}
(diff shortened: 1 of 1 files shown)


cmd/runtimecfg/node-ip.go
```diff
@@ -145,6 +145,20 @@ func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool) (chosen []net.IP
 		if len(vips) > 0 {
 			chosen, err = utils.AddressesRouting(vips, utils.ValidNodeAddress)
 			if len(chosen) > 0 || err != nil {
+
+				// If using IPv6, verify that the choosen address isn't tentative
+				// i.e. we can actually bind to it
+				if net.IPv6len == len(chosen[0]) {
+					_, err := net.Listen("tcp", "["+chosen[0].String()+"]:10010")
```
hardys, line 152: Binding to a hard-coded port seems a little hacky - although we know that the way things are deployed on startup this should run prior to `crio.service`, this would fail if someone runs the node-ip check manually after it's running, right?

I'm wondering if we should either use a random high port, or ideally check the connection flags e.g can we determine if the state is/isn't tentative via the golang rtnetlink bindings?

https://pkg.go.dev/github.com/vishvananda/go-netlink/rtnetlink/addr#Flags

derekhiggins: I can remove the port number and a random on is used, I didn't do this as it may pick a port another service wants, but its so short lived I guess it should be ok

I can Look into checking the flags

cybertron:
lgtm pending ipv6 job pass
/bugzilla refresh

Note that it is important for this to work after crio has already started because the node-ip logic is also used in the resolv-prepender script that makes sure our local DNS server is first in resolv.conf. I'm a little confused how the ipv6 job passed on the previous version of this.

cybertron:
/lgtm

Looks like this will unblock ipv6 ci and it's not an unreasonable thing to check, so let's go with it.

PR #183, opened 2022-07-05
RFE-2962: Set interface name from node-ip-configuration to configure-ovs

tsorya:
node-ip-configuration will set env file for configure-ovs when it will
set kubelet ip, this will allow to configure ovs with the bridge on the
same interface that kubelet will use

unitests were not fixed, will do it if we decide this idea is good to go

https://github.com/openshift/machine-config-operator/pull/3222

### DIFF
--- a/cmd/runtimecfg/node-ip.go
+++ b/cmd/runtimecfg/node-ip.go
@@ -16,8 +16,9 @@ import (
 )
 
 const (
-	kubeletSvcOverridePath = "/etc/systemd/system/kubelet.service.d/20-nodenet.conf"
-	crioSvcOverridePath    = "/etc/systemd/system/crio.service.d/20-nodenet.conf"
+	kubeletSvcOverridePath          = "/etc/systemd/system/kubelet.service.d/20-nodenet.conf"
+	crioSvcOverridePath             = "/etc/systemd/system/crio.service.d/20-nodenet.conf"
+	ovsConfigurationSvcOverridePath = "/etc/systemd/system/ovs-configuration.service.d//20-nodenet.conf"
 )
 
 var retry, preferIPv6 bool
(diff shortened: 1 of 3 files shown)


cmd/runtimecfg/node-ip.go
```diff
@@ -16,8 +16,9 @@ import (
 )
 
 const (
-	kubeletSvcOverridePath = "/etc/systemd/system/kubelet.service.d/20-nodenet.conf"
-	crioSvcOverridePath    = "/etc/systemd/system/crio.service.d/20-nodenet.conf"
+	kubeletSvcOverridePath          = "/etc/systemd/system/kubelet.service.d/20-nodenet.conf"
+	crioSvcOverridePath             = "/etc/systemd/system/crio.service.d/20-nodenet.conf"
+	ovsConfigurationSvcOverridePath = "/etc/systemd/system/ovs-configuration.service.d//20-nodenet.conf"
```
romfreiman, line 21: Extra /

cmd/runtimecfg/node-ip.go
```diff
@@ -129,27 +130,52 @@ func set(cmd *cobra.Command, args []string) error {
...
-	cOverrideContent := fmt.Sprintf("[Service]\nEnvironment=\"CONTAINER_STREAM_ADDRESS=%s\"\n", chosenAddresses[0])
+	cOverrideContent := fmt.Sprintf("[Service]\nEnvironment=\"CONTAINER_STREAM_ADDRESS=%s\"\n", chosenAddresses[0].Address)
 	log.Infof("Writing CRI-O service override with content %s", cOverrideContent)
 	_, err = cOverride.WriteString(cOverrideContent)
 	if err != nil {
 		return err
 	}
+
+	// OVN
+	// Set kubelet interface in case it is not br-ex that means that it was already  configured and node is rebooting
```
dougsland, line 141: Extra space in "already  configured"

cmd/runtimecfg/node-ip.go
```diff
@@ -129,27 +130,52 @@ func set(cmd *cobra.Command, args []string) error {
...
+	// in that case don't change the previous written interface
+	if chosenAddresses[0].LinkName != "br-ex" {
```
dougsland, line 143: "br-ex" a candidate for const?

cmd/runtimecfg/node-ip.go
```diff
@@ -129,27 +130,52 @@ func set(cmd *cobra.Command, args []string) error {
...
+		ovsConfigurationDir := filepath.Dir(ovsConfigurationSvcOverridePath)
+		err = os.MkdirAll(ovsConfigurationDir, 0755)
```
dougsland, line 145: Should we check if the dir exists and avoid getting an error creating it?

cmd/runtimecfg/node-ip.go
dougsland, line 143: from line 144 until 160 could be converted to a helper function?

dougsland (commented):
The patch looks sane, few comments only.

tsorya:
it was decided to go with ip address and not setting interface

PR #185, opened 2022-07-06
Adding node ip hint for all who want to use it

tsorya:


tsorya:
@cybertron this is what you meant in https://github.com/openshift/machine-config-operator/pull/3222 ?

dougsland (commented):
it seems a good patch but it's in Draft mode. Are you planing to update it?

tsorya:
@dougsland waited for enhancement to be pushed

cmd/runtimecfg/node-ip.go
```diff
@@ -135,6 +136,27 @@ func set(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+	log.Infof("Opening node ip override path %s", nodeIpFile)
+	nodeIpOverride, err := os.Create(nodeIpFile)
+	if err != nil {
+		return err
+	}
+	defer nodeIpOverride.Close()
+
+	nodeIpOverrideContent := fmt.Sprintf("NODE_IP=%s\n", nodeIP)
```
cybertron, line 153: I would prefer to just write the IP, rather than make this a systemd service-specific format. configure-ovs can still easily extract the IP from the file and it makes it easier for anything else that needs to consume this.

The goal here was to stop writing service-specific override files and just provide a single source of truth for the node ip.

cybertron (changes requested):

cmd/runtimecfg/node-ip.go
```diff
@@ -91,53 +94,78 @@ func set(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+	for _, ip := range chosenAddresses[0:1] {
+		if utils.IsIPv6(ip) && !fileExists(nodeIpIpV6File) {
+			err = writeToFile(nodeIpIpV6File, chosenAddresses[0].String())
+			if err != nil {
+				return err
+			}
+		} else if !utils.IsIPv6(ip) && !fileExists(nodeIpIpV4File) {
+			err = writeToFile(nodeIpIpV4File, chosenAddresses[0].String())
```
cybertron, line 130: Same thing here.

cmd/runtimecfg/node-ip.go
```diff
@@ -91,53 +94,78 @@ func set(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return err
+	}
+
+	// node ip hint for all other services
+	err = writeToFile(nodeIpFile, nodeIP)
+	if err != nil {
+		return err
+	}
+	for _, ip := range chosenAddresses[0:1] {
```
cybertron, line 123: Somewhat confusingly, this doesn't include the 0 and 1 items. It excludes the upper index, so it looks like this needs to be chosenAddresses[0:2] (or just [:2], which seems to be equivalent).

cybertron (changes requested):
Just a couple of things from my test in a dual stack env. Otherwise lgtm.

cmd/runtimecfg/node-ip.go
tsorya: sorry, missed that somehow

cmd/runtimecfg/node-ip.go
tsorya: totally right, thank you

cmd/runtimecfg/node-ip.go
```diff
@@ -91,53 +94,79 @@ func set(cmd *cobra.Command, args []string) error {
...
 	if err != nil {
 		return err
 	}
-	defer cOverride.Close()
 
-	cOverrideContent := fmt.Sprintf("[Service]\nEnvironment=\"CONTAINER_STREAM_ADDRESS=%s\"\n", chosenAddresses[0])
-	log.Infof("Writing CRI-O service override with content %s", cOverrideContent)
-	_, err = cOverride.WriteString(cOverrideContent)
+	for i := 0; i < len(chosenAddresses) && i < 2; i++ {
+		if utils.IsIPv6(chosenAddresses[i]) && !fileExists(nodeIpIpV6File) {
```
rbbratta, line 125: Is there a race condition between the `os.Stat` and `os.Create`?  I don't think it is atomic.   

We could maybe just check for `EEXIST` from the `os.Create` error using `errors.Is(err, fs.ErrExist)`

cmd/runtimecfg/node-ip.go
```diff
@@ -91,53 +94,79 @@ func set(cmd *cobra.Command, args []string) error {
...
+			err = writeToFile(nodeIpIpV6File, chosenAddresses[i].String())
+			if err != nil {
+				return err
+			}
+		} else if !utils.IsIPv6(chosenAddresses[i]) && !fileExists(nodeIpIpV4File) {
```
rbbratta, line 130: is it expensive to check if the address is IPv6 twice?

tsorya: nope, it is actually checking the string

cmd/runtimecfg/node-ip.go
tsorya: i will remove file check, will just set boolean will be more obvious i assume

cmd/runtimecfg/node-ip.go
tsorya: but agree that there is no need for it

cmd/runtimecfg/node-ip.go
```diff
@@ -91,47 +94,68 @@ func set(cmd *cobra.Command, args []string) error {
...
+
+	ipv6Created, ipv4Created := false, false
+	for i := 0; i < len(chosenAddresses) && i < 2; i++ {
+		if utils.IsIPv6(chosenAddresses[i]) && !ipv6Created {
+			err = writeToFile(nodeIpIpV6File, chosenAddresses[i].String())
+			if err != nil {
+				return err
+			}
+			ipv6Created = true
+		} else if !ipv4Created {
```
cybertron, line 132: This makes me a little nervous. If we ever got two ipv6 chosenAddresses, it could potentially result in us writing an ipv6 address to the ipv4 file. To avoid checking IsIPv6 twice (which, honestly, I think is premature optimization. This is a service that runs once per boot and is not performance critical code), I think we need that to be an outer check and then within the cases for it we do the ipv*Created checks. Something like:

```
if isipv6 {
  if !ipv6Created {
    create file
  }
} else {
  if !ipv4Created {
     create file
  }
}
```

Note that I don't think it's currently possible to get two ipv6 addresses in chosenAddresses, but if I'm wrong or that changes in the future I'd prefer to have this code handle it correctly.

cybertron (changes requested):

cmd/runtimecfg/node-ip.go
tsorya: This is actually a bug from my side . We should check if nit ipv6 , this is why it was there from the start. Sorry

cmd/runtimecfg/node-ip.go
```diff
@@ -142,7 +166,7 @@ func checkAddressUsable(chosen []net.IP) (err error) {
 	// If using IPv6, verify that the choosen address isn't tentative
 	// i.e. we can actually bind to it
 	if len(chosen) > 0 && net.IPv6len == len(chosen[0]) {
-		_, err = net.Listen("tcp", "["+chosen[0].String()+"]:")
+		_, err = net.Listen("tcp", "["+chosen[0].String()+"]:10010")
```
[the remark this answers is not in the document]
tsorya: Added it as test and forgot to remove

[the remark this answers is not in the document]
cybertron: Okay, I think we should remove it so we aren't adding another required port. Since ci has passed on this version, do you want to do that as a followup?

cmd/runtimecfg/node-ip.go
```diff
@@ -142,7 +166,7 @@ func checkAddressUsable(chosen []net.IP) (err error) {
 	// If using IPv6, verify that the choosen address isn't tentative
 	// i.e. we can actually bind to it
 	if len(chosen) > 0 && net.IPv6len == len(chosen[0]) {
-		_, err = net.Listen("tcp", "["+chosen[0].String()+"]:")
+		_, err = net.Listen("tcp", "["+chosen[0].String()+"]")
```
cybertron, line 169: You may need to keep the : in the address string. I'm not positive, but the docs suggest that's the correct syntax.

cybertron:
/lgtm
/approve

Checked locally in a dual stack env and this correctly populated all of the files. We should be good to go here.

PR #186, opened 2022-07-19
RFE-2962: Set interface name from node-ip-configuration to configure-ovs

tsorya:
node-ip-configuration will set env file for configure-ovs when it will
set kubelet ip, this will allow to configure ovs with the bridge on the
same interface that kubelet will use

### DIFF
--- a/cmd/runtimecfg/node-ip.go
+++ b/cmd/runtimecfg/node-ip.go
@@ -18,6 +18,8 @@ import (
 const (
 	kubeletSvcOverridePath = "/etc/systemd/system/kubelet.service.d/20-nodenet.conf"
 	crioSvcOverridePath    = "/etc/systemd/system/crio.service.d/20-nodenet.conf"
+	interfaceFile          = "/run/nodeip-configuration/interface"
+	primaryFile            = "/run/nodeip-configuration/primary-ip"
 )
 
 var retry, preferIPv6 bool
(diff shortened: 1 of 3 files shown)


cmd/runtimecfg/node-ip.go
```diff
@@ -129,27 +131,70 @@ func set(cmd *cobra.Command, args []string) error {
...
 	_, err = cOverride.WriteString(cOverrideContent)
 	if err != nil {
 		return err
 	}
+
+	// Set interface in case it is not br-ex that means that it was already configured and node is rebooting
+	// in that case don't change the previous written interface
+	if chosenAddresses[0].LinkName != "br-ex" {
+		interfaceDir := filepath.Dir(interfaceFile)
+		err = os.MkdirAll(interfaceDir, 0755)
```
dougsland, line 145: should we check if the interfaceDir exists before trying to create?

cmd/runtimecfg/node-ip.go
```diff
@@ -129,27 +131,70 @@ func set(cmd *cobra.Command, args []string) error {
...
+	cOverrideContent := fmt.Sprintf("[Service]\nEnvironment=\"CONTAINER_STREAM_ADDRESS=%s\"\n", chosenAddresses[0].Address)
 	log.Infof("Writing CRI-O service override with content %s", cOverrideContent)
 	_, err = cOverride.WriteString(cOverrideContent)
 	if err != nil {
 		return err
 	}
+
+	// Set interface in case it is not br-ex that means that it was already configured and node is rebooting
+	// in that case don't change the previous written interface
+	if chosenAddresses[0].LinkName != "br-ex" {
```
dougsland, line 143: br-ex seems a good candidate for const ?

dougsland (commented):
The patch is draft but it's in my queue for review, looks good but I have few comments.

dougsland:
Similar to https://github.com/openshift/baremetal-runtimecfg/pull/183 ?

tsorya:
it was decided to go with ip address and not setting interface

PR #188, opened 2022-08-24
OCPBUGS-669: Apply ipv6 bind check to non-VIP case too

cybertron:
In #181 we fixed a problem where ipv6 addresses are initially in a
state that is not usable by services. However, this was only done
for the code path where VIPs are in use, so it only applies to IPI
deployments that are not using remote workers. It turns out this is
also happening on non-IPI deployments (and likely for IPI deployments
with remote workers), so we need to run the same check on that code
path.

### DIFF
--- a/cmd/runtimecfg/node-ip.go
+++ b/cmd/runtimecfg/node-ip.go
@@ -138,6 +138,19 @@ func set(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func checkAddressUsable(chosen []net.IP) (err error) {
+	// If using IPv6, verify that the choosen address isn't tentative
+	// i.e. we can actually bind to it
+	if len(chosen) > 0 && net.IPv6len == len(chosen[0]) {
+		_, err = net.Listen("tcp", "["+chosen[0].String()+"]:")
+		if err != nil {
+			log.Errorf("Chosen node IP is not usable")
+			return err
+		}
+	}
+	return err
+}
+
 func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool) (chosen []net.IP, err error) {
 	// Enable debug logging in utils package
 	utils.SetDebugLogLevel()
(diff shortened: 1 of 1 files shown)


cmd/runtimecfg/node-ip.go
```diff
@@ -138,33 +138,52 @@ func set(cmd *cobra.Command, args []string) error {
...
-						log.Errorf("Chosen node IP is not usable")
-						if !retry {
-							return nil, fmt.Errorf("Failed to find node IP")
-						}
-						time.Sleep(time.Second)
-						continue
+				if err == nil {
+					err = checkAddressUsable(chosen)
+				}
+				if err != nil {
```
dougsland, line 164: I am wondering if this logic should go to below line 147 and the line 161 inside a if like: if len(choosen) > 0 && err == nil {

cmd/runtimecfg/node-ip.go
```diff
@@ -138,33 +138,52 @@ func set(cmd *cobra.Command, args []string) error {
...
+					time.Sleep(time.Second)
+					continue
 				}
 				return chosen, err
 			}
 		}
 		if len(chosen) == 0 {
 			chosen, err = utils.AddressesDefault(preferIPv6, utils.ValidNodeAddress)
 			if len(chosen) > 0 || err != nil {
+				if err == nil {
```
dougsland, line 177: This logic looks similar to line 164 and so on. Should we convert to a helper function to avoid dup code?

dougsland (commented):
The patch looks good, I have few comments.

cmd/runtimecfg/node-ip.go
cybertron: Initially I was going to do this, but because there's a loop flow control statement I don't think we can put this block in the function.

cmd/runtimecfg/node-ip.go
cybertron: Same problem. Things get weird if we move more of this logic into the function because we end up needing some way to communicate flow control back into the loop here. It is a bit duplicated, but I think the code is more clear than if we try to deduplicate this any further.

dougsland:
@cybertron it seems missing a bugzilla for merge or should we ignore the label?

cybertron:
I opened a Jira clone of the bz so we can track the OCP-side fix for this while NetworkManager looks at the more general solution. This should be good to go.

PR #190, opened 2022-08-31
Add note about removing compatibility code

cybertron:
The coercion of single vips to a list will no longer be necessary
once we remove the single vip interface to runtimecfg. Make a note
of that in the comments on the code.


dougsland:
@cybertron can we please squash the patches for easier review?

PR #191, opened 2022-09-01
OCPBUGS-669: Empty chosen list when retrying ip selection

cybertron:
Because chosen is function-scoped, it is persistent over multiple
iterations of the loop. If the first iteration fails and we retry,
the second time around we skip the AddressesDefault case because
chosen is no longer empty. This means the retries will never succeed.

If we just re-initialize the chosen list before retrying then it
should get repopulated the next time through and the retry mechanism
will work as expected.


dougsland (approved):
/lgtm

PR #197, opened 2022-10-04
Run go fmt

creydr:
#196 fails due to format issues in node.go. This PR addresses it and formats the go files.


cybertron:
/lgtm

Huh, how did that happen? We've been running go fmt in CI on this repo for years. o.O

PR #198, opened 2022-10-05
Add keepalived healthcheck logic to monitor

cybertron:
When a cluster gets into a bad state and the api gets overloaded to the point where we start getting a lot of errors from our healthchecks, the current keepalived logic tends to flap and cause unnecessary failovers. This compounds the problem with apiserver stability because a failover forces every service to reconnect.

The rise and fall values of the healthcheck scripts are intended to help with this and they do, but there are still some situations where they are insufficient. By moving the logic into runtimecfg we are able to do more sophisticated checks, including dynamic rise and fall values based on which node holds the VIP. This significantly reduces the probability of failovers in a flaky cluster but should not have much effect on a normal functioning cluster.

### DIFF
--- a/pkg/monitor/dynkeepalived.go
+++ b/pkg/monitor/dynkeepalived.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"syscall"
 	"time"
+	"crypto/tls"
 
 	"gopkg.in/yaml.v2"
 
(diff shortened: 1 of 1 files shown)


mdbooth (changes requested):
This adds complexity to the health check but it still relies on a healthy response from kube-apiserver which is not relevant to keepalived. It can still cause a loadbalancer failure due to non-deterministic failures from kube-apiserver. We've just made these harder to reason about.

Think of this from the POV of any other cloud provider. This would be akin to deleting and replacing an AWS loadbalancer because the kube apiserver wasn't working: it doesn't make sense. AWS internally might replace the loadbalancer if it failed, and this is the role that keepalived is providing for us. But AWS would not move the loadbalancer just because its backends were unhealthy. A datacenter wouldn't replace a switch just because the connected servers weren't healthy. These things aren't relevant.

I worry that we're just making our failure modes more complex.

PR #199, opened 2022-10-18
OCPBUGS-2512: Improve IP address sort order for interface selection

andreaskaris:
How to test:
~~~
launch openshift/baremetal-runtimecfg#199 metal,ipv6
~~~

### DIFF
--- a/pkg/utils/addresses.go
+++ b/pkg/utils/addresses.go
@@ -204,9 +204,10 @@ func AddressesDefault(preferIPv6 bool, af AddressFilter) ([]net.IP, error) {
 }
 
 type FoundAddress struct {
-	Address   net.IP
-	Priority  int
-	LinkIndex int
+	Address         net.IP
+	Priority        int
+	LinkIndex       int
+	GatewayOnSubnet bool
 }
 
 func addressesDefaultInternal(preferIPv6 bool, af AddressFilter, getAddrs addressMapFunc, getRouteMap routeMapFunc) ([]net.IP, error) {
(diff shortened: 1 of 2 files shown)


pkg/utils/addresses.go
```diff
@@ -230,20 +248,32 @@ func addressesDefaultInternal(preferIPv6 bool, af AddressFilter, getAddrs addres
...
 	}
 
-	// Sort addresses into a stable order, based on default route priority and link
+	// Sort addresses into a stable order, based on:
+	// a) default route priority
+	// b) link
+	// c) IP address family
+	// d) The default route matching the IP CIDR
+	// e) The address being public
 	// index. Otherwise the order of the addresses we return may change if an address
```
cybertron, line 265: Nit: The "index." here is a leftover from the now-removed comment.

pkg/utils/addresses.go
```diff
@@ -94,6 +101,16 @@ func ValidNodeAddress(address netlink.Addr) bool {
 		return false
 	}
 
+	// Ignore the IPv6 OVNK masquerade subnet. We should never see an error here as this is a constant.
+	// If we get an error, panic.
+	_, netip, err := net.ParseCIDR(V6MasqueradeSubnet)
+	if err != nil {
+		panic(err)
+	}
+	if netip.Contains(address.IP) {
```
cybertron, line 110: I still don't think we can do this unconditionally. The masquerade subnet is a valid address range in non-OVNK CNIs.

andreaskaris: I removed it altogether for this PR

pkg/utils/addresses.go
andreaskaris: Changed + renamed one of the variables

pkg/utils/addresses.go
```diff
@@ -230,20 +231,30 @@ func addressesDefaultInternal(preferIPv6 bool, af AddressFilter, getAddrs addres
...
 	}
 
-	// Sort addresses into a stable order, based on default route priority and link
-	// index. Otherwise the order of the addresses we return may change if an address
-	// moves to a bridge (for example).
+	// Sort addresses into a stable order, based on:
+	// a) default route priority
+	// b) link
+	// c) IP address family
+	// d) The gatway being on the IP's CIDR
```
tsorya, line 246: nit: gateway

tsorya:
lgtm as for me

pkg/utils/addresses.go
andreaskaris: fixed, me--

cybertron:
This lgtm. Do we want anyone else to sign off before we merge?

andreaskaris:
@jcaamano

andreaskaris:
> /lgtm
> 
> I am surprised though with the prioritization by link index that was already there. What's that supposed to represent?

From the old comment (which I removed):
~~~
// Sort addresses into a stable order, based on default route priority and link
	// index. Otherwise the order of the addresses we return may change if an address
	// moves to a bridge (for example).
~~~

andreaskaris:
@cybertron I think this is ready to merge, then!

cybertron:
The link index sorting comes from https://github.com/openshift/baremetal-runtimecfg/pull/130 I think that may have actually been unrelated to the address moving to a bridge, but it got included in the same comment because that's what was there.

PR #200, opened 2022-10-18
OCPBUGS-2512: If primary ip address was already created no need to choose new ip

tsorya:
OCPBUGS-2512: If primary ip address was already created, it means that nodeip-configuration has run already and no need to
choose new ip, we should leave same configuration as we already set


cmd/runtimecfg/node-ip.go
```diff
@@ -83,6 +83,12 @@ func show(cmd *cobra.Command, args []string) error {
 }
 
 func set(cmd *cobra.Command, args []string) error {
+	// If primary ip address was already created, it means that nodeip-configuration has run already and no need to
+	// choose new ip, we should leave same configuration as we already set
+	if _, err := os.Stat(nodeIpFile); err == nil {
```
rbbratta, line 88: Do we need to check for a zero length file or lack of read permissions?

`os.Stat` tends to be a poor choice, it seems better to test what we actual need, which is to read file that is not empty, or even better that contains a parseable IP.

tsorya: Sounds reasonable

tsorya: But lets see if this PR is relevant all, @cybertron ?

tsorya:
/hold
requires rewriting

cmd/runtimecfg/node-ip.go
andreaskaris: So I think we agreed that we still need this one here, correct? So change this to reading file + parsing IP address?

tsorya:
@cybertron if it matches what you asked for, i will add tests

cybertron (changes requested):
I think this will be problematic for dual stack vips. If we wait for https://github.com/openshift-metal3/dev-scripts/pull/1467 to merge we'll be able to verify that using ci (assuming we have a dual stack job for this repo, I guess).

cybertron (commented):
/lgtm

We'll need to retest once the MCO patch merges, but this is working for me locally and I think it does what we want.

cmd/runtimecfg/node-ip.go
tsorya: yep

cmd/runtimecfg/node-ip.go
```diff
@@ -83,6 +84,13 @@ func show(cmd *cobra.Command, args []string) error {
 }
 
 func set(cmd *cobra.Command, args []string) error {
+	// If primary ip address was already created, it means that nodeip-configuration has run already and no need to
+	// choose new ip, we should leave same configuration as we already set
+	if ip, err := config.GetIpFromFile(nodeIpFile); err == nil {
```
dougsland, line 89: nit: What happens if we get err here, just ignore?

pkg/config/net.go
```diff
@@ -2,12 +2,48 @@ package config
...
+// In case of error return empty string
+func GetIpFromFile(filePath string) (net.IP, error) {
+	b, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		log.WithError(err).Infof("Failed to read ip from file %s", filePath)
+		return nil, err
+	}
+	ip := net.ParseIP(string(b))
+	if ip == nil {
+		msg := fmt.Sprintf("Failed to parse ip from file %s", filePath)
```
dougsland, line 27: nit: other places are using `log.WithError()` which looks like avoid explicit fmt.Sprint() if I got that correctly. 
Should we use it here too?

dougsland:
@cybertron it shows you requested changes, still valid?

cmd/runtimecfg/node-ip.go
tsorya: Yep. Failed to read file go to the main logic

pkg/config/net.go
tsorya: There is no error here to print parse ip function so I create new one a don't want to write same string  twice

cmd/runtimecfg/node-ip.go
dougsland: Adding `err != nil` and `return err` early seems okay too.

pkg/config/net.go
dougsland: Just to clarify, I was referring to line `22` as reference.

cmd/runtimecfg/node-ip.go
tsorya: I don't actually think this will be right in that case. There is no point of stopping run because if this error.
We should continue running till the end as this can actually fix this error by , lets say, writing ip to the file that somehow was written wrong previously or touched by user for unknown reason.
I really thinks that code in such critical service should be resilient to errors as much as possible and we should continue running till we can, we will not win anything in falling 0.5 seconds before.

dougsland: Sure, that's okay for me.

PR #201, opened 2022-10-20
OCPBUGS-2994: Adding timeout for kubeapi calls for keepalived monitor

tsorya:
OCPBUGS-2994: Keepalived monitor stuck for long period of time on kube-api call while installing
While installing kube-api server is not stable and keepalived-monitor can stuck while trying to get nodes

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -7,12 +7,17 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/sirupsen/logrus"
 )
+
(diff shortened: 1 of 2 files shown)


pkg/utils/utils.go
```diff
@@ -115,3 +118,18 @@ func GetFileMd5(filePath string) (string, error) {
...
+	config, err := clientcmd.BuildConfigFromFlags(kubeApiServerUrl, kubeconfigPath)
+	if err != nil {
+		log.WithFields(logrus.Fields{
+			"err": err,
+		}).Info("Failed to get client config")
+		return nil, err
+	}
+	// Kubeapi can be not stable on installation process
+	// and we should free connection in case it was stuck
+	config.Timeout = 30 * time.Second
```
dougsland, line 133: nit: how about a constant instead of 30 here  ?

cybertron:
/approve cancel

Should wait for the metal-ipi job.

PR #205, opened 2022-11-10
node-ip: no need to log error in infinite loop

dougsland:
In case the node has link down or other situation
preveting the ip address be assigned DO NOT print
error message in the infite loop.


dougsland:
Closing as I will prepare another patch.

/close

PR #207, opened 2022-11-23
OPNET-133: Support remote worker in onprem installations

tsorya:
[OPNET-133](https://issues.redhat.com//browse/OPNET-133): Support remote worker in onprem installations
1. Create file that will be used by keepalived in case chosen ip is
   different from vips (remote worker)
2. Set remote worker label that will be used by ingress

### DIFF
--- a/cmd/runtimecfg/node-ip.go
+++ b/cmd/runtimecfg/node-ip.go
@@ -17,11 +17,13 @@ import (
 )
 
 const (
-	kubeletSvcOverridePath = "/etc/systemd/system/kubelet.service.d/20-nodenet.conf"
-	nodeIpFile             = "/run/nodeip-configuration/primary-ip"
-	nodeIpIpV6File         = config.NodeIpIpV6File
-	nodeIpIpV4File         = config.NodeIpIpV4File
-	crioSvcOverridePath    = "/etc/systemd/system/crio.service.d/20-nodenet.conf"
+	kubeletSvcOverridePath   = "/etc/systemd/system/kubelet.service.d/20-nodenet.conf"
+	nodeIpFile               = "/run/nodeip-configuration/primary-ip"
+	nodeIpIpV6File           = config.NodeIpIpV6File
+	nodeIpIpV4File           = config.NodeIpIpV4File
+	nodeIpNotMatchesVipsFile = "/run/nodeip-configuration/remote-worker"
+	crioSvcOverridePath      = "/etc/systemd/system/crio.service.d/20-nodenet.conf"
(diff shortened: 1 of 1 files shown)


cmd/runtimecfg/node-ip.go
```diff
@@ -21,7 +21,9 @@ const (
 	nodeIpFile             = "/run/nodeip-configuration/primary-ip"
 	nodeIpIpV6File         = config.NodeIpIpV6File
 	nodeIpIpV4File         = config.NodeIpIpV4File
+	nodeIpNotMatchesVipsFile  = "/run/nodeip-configuration/notMatchesVips"
```
cybertron, line 24: Nit: For consistency I would prefer to call the file "not-matches-vips".

cybertron (commented):
Overall this looks good to me, but a recent discussion about some of the new validations in the installer has brought up a potential issue here. Apparently non-baremetal platforms were allowing VIPs to be outside the machine network, which means that even non-remote workers might not match the node ip subnet. I don't understand how this even works, but apparently it does.

I'm not entirely sure how to handle that. It breaks multiple assumptions in our networking architecture and I think it only works because of the remote worker fallback in the ip selection code. We may need to make this optional so anyone deploying that way can still have keepalived on nodes that don't match the VIP.

cmd/runtimecfg/node-ip.go
```diff
@@ -21,7 +21,9 @@ const (
 	nodeIpFile             = "/run/nodeip-configuration/primary-ip"
 	nodeIpIpV6File         = config.NodeIpIpV6File
 	nodeIpIpV4File         = config.NodeIpIpV4File
+	nodeIpNotMatchesVipsFile  = "/run/nodeip-configuration/not-matches-vips"
```
cybertron, line 24: Sorry for changing my mind on this, but I think I'd mildly prefer to call the file "remote-worker" instead of "not-matches-vips". It reads better and I think it communicates the intent of the file more clearly.

cybertron (commented):
/lgtm

Two minor things: The go fmt failure will need to be fixed, and while we're doing that I would mildly prefer to rename the file again (sorry!). This isn't user-facing so I'm not going to block it over something that trivial, but since another revision is needed anyway hopefully it isn't too big a deal.

cybertron:
/hold

Oops, I didn't think this would go with failing jobs. It can't pass until the formatting error is fixed so there's no point re-running jobs.

cmd/runtimecfg/node-ip.go
tsorya: wanted to change it myself, forgot though. Yeah totally agree

PR #212, opened 2022-12-07
OPNET-192: update centos version

dougsland:
Update Dockerfile for most recent CentOS Stream (9).

Signed-off-by: Douglas Schilling Landgraf <dlandgra@redhat.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ COPY . .
 RUN mkdir build
 RUN GO111MODULE=on go build --mod=vendor -o build ./cmd/...
 
-FROM centos:stream8
+FROM centos:stream9
 
 RUN yum install -y dhcp-client diffutils && yum clean all
 


dougsland:
Also tested against OCP 4.11 and 4.12 everything worked.

mandre (commented):
The `verify-api-int-resolvable` is effectively [running from the baremetal-runtimecfg image](https://github.com/openshift/machine-config-operator/blob/0357c4b/templates/master/00-master/on-prem/files/haproxy.yaml#L31-L32) and the container seems to be [working fine](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_baremetal-runtimecfg/212/pull-ci-openshift-baremetal-runtimecfg-master-e2e-metal-ipi/1600555505170780160/artifacts/e2e-metal-ipi/gather-extra/artifacts/pods/openshift-kni-infra_haproxy-master-0_verify-api-int-resolvable.log) with this change.

However, while useful to bump dependencies, this just shows that the centos-9 based baremetal-runtimecfg image can execute `oc` from a RHEL8-based RHCOS.

Is this Dockerfile the source for baremetal-runtimecfg image that will be in the OCP payload, or is it only used in CI?
I'd like to understand if it's correct to reference [OCPBUGS-3988](https://issues.redhat.com//browse/OCPBUGS-3988) here.

dougsland:
> The `verify-api-int-resolvable` is effectively [running from the baremetal-runtimecfg image](https://github.com/openshift/machine-config-operator/blob/0357c4b/templates/master/00-master/on-prem/files/haproxy.yaml#L31-L32) and the container seems to be [working fine](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_baremetal-runtimecfg/212/pull-ci-openshift-baremetal-runtimecfg-master-e2e-metal-ipi/1600555505170780160/artifacts/e2e-metal-ipi/gather-extra/artifacts/pods/openshift-kni-infra_haproxy-master-0_verify-api-int-resolvable.log) with this change.
> 
> However, while useful to bump dependencies, this just shows that the centos-9 based baremetal-runtimecfg image can execute `oc` from a RHEL8-based RHCOS.
> 
> Is this Dockerfile the source for baremetal-runtimecfg image that will be in the OCP payload, or is it only used in CI? I'd like to understand if it's correct to reference [OCPBUGS-3988](https://issues.redhat.com//browse/OCPBUGS-3988) here.

My understanding is that CI uses Dockerfile.openshift. I was expecting that `master` branch shouldn't be a problem replacing the stream from 8 to 9.

dougsland:
Adding @dollierp  in the loop.

mandre:
It seems like [CI is using Dockerfile.openshift](https://github.com/openshift/release/blob/e4bd338/ci-operator/config/openshift/baremetal-runtimecfg/openshift-baremetal-runtimecfg-master.yaml#L17), so the change was apparently not tested in CI. Not sure where Dockerfile is used.

cybertron:
This is only used for local builds (like in dev-scripts). This is probably fine since 4.13 will be based on 9, but it doesn't fix the problem in actual release builds.

I'm actually wondering if we should just change how we check for api-int. There's no particular reason we need to call oc, we could just curl api-int.cluster-domain/readyz which would be a lot lighter and wouldn't have this problem.

cybertron:
I changed the reference on this to point at the RHCOS 9 epic instead of the bug. We should probably wait to merge it until the regular OpenShift base images are 9-based too.

PR #215, opened 2023-01-12
OCPBUGS-6004: filter ovn fd69::2 ip from list of ips in case ovn

tsorya:
Adding network type in order to be able to filter ovn ipv6 ip

### DIFF
--- a/pkg/utils/addresses.go
+++ b/pkg/utils/addresses.go
@@ -100,6 +100,16 @@ func ValidNodeAddress(address netlink.Addr) bool {
 	return true
 }
 
+// ValidOVNNodeAddress returns true if the address is suitable for a node's primary IP
+// and is not fd69::2
+// we intentionally don't filter ipv4 because it isn't necessary.
+func ValidOVNNodeAddress(address netlink.Addr) bool {
+	if IsIPv6(address.IP) && address.IP.String() == "fd69::2" {
+		return false
+	}
+	return ValidNodeAddress(address)
+}
+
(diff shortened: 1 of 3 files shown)


tsorya:
Looks good

Jan 16 15:33:25 02-00-00-4f-27-17 bash[18853]: time="2023-01-16T15:33:25Z" level=debug msg="Ignoring filtered address fd69::2/125"

pkg/utils/addresses.go
```diff
@@ -241,6 +251,7 @@ func addressesDefaultInternal(preferIPv6 bool, af AddressFilter, getAddrs addres
 			})
 		}
 	}
+	fmt.Println(addrs)
```
cybertron, line 254: Leftover debug statements? We typically use the log module for messages we want to keep.

pkg/utils/addresses.go
```diff
@@ -277,6 +288,7 @@ func addressesDefaultInternal(preferIPv6 bool, af AddressFilter, getAddrs addres
 			foundv4 = true
 		}
 	}
+	fmt.Println(matches)
```
cybertron, line 291: Ditto

cybertron (commented):
LGTM modulo a couple of printfs that should either be removed or converted to log calls.

pkg/utils/addresses.go
tsorya: Yep, removing

cybertron (commented):
/lgtm

cybertron:
/hold

We should resolve the bug situation with this.

PR #218, opened 2023-01-30
OPNET-197: Extend logic for detecting Node IP

mkowalski:
When generating keepalived.conf we are relying on the logic to gather IPs of all the cluster nodes for the IP stack used by the specific VIP. This logic currently relies only on the addresses reported as part of Node.Status.Addresses.

In some scenarios it may be that the node is not reporting all its IPs via kubelet but still have those available. If we detect such a scenario (e.g. kubelet reporting only IPv4, but VIP being IPv6), we will check for Node annotations created by OVN as those use different source of truth so kubelet not reporting IPs is not affecting it.

The newly introduced behaviour is just a fallback in case Node.Status.Addresses does not contain an IP of a requested stack, therefore not changing the behaviour for currently working scenarios.

Contributes-to: [OPNET-197](https://issues.redhat.com//browse/OPNET-197)

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -3,6 +3,7 @@ package config
 import (
 	"bufio"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -291,7 +292,10 @@ func IsUpgradeStillRunning(kubeconfigPath string) (bool, error) {
 	return true, nil
 }
 
-func GetIngressConfig(kubeconfigPath string, filterIpType string) (ingressConfig IngressConfig, err error) {
+func GetIngressConfig(kubeconfigPath string, vips []string) (IngressConfig, error) {
+	var machineNetwork string
+	var ingressConfig IngressConfig
+
 	config, err := utils.GetClientConfig("", kubeconfigPath)
 	if err != nil {
 		return ingressConfig, err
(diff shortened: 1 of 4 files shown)


pkg/utils/network.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	return split[0], split[1], nil
+}
+
+func IsIPv4Addr(ip string) bool {
+	return strings.Contains(ip, ".") && net.ParseIP(ip) != nil
+}
+
+func IsIPv6Addr(ip string) bool {
```
cybertron, line 23: Note that we already have an IsIPv6 function: https://github.com/openshift/baremetal-runtimecfg/blob/2905c0425f6e8f2952d366b130f5583da04e1e6d/pkg/utils/addresses.go#L131

pkg/config/node.go
```diff
@@ -483,15 +549,18 @@ func getSortedBackends(kubeconfigPath string, readFromLocalAPI bool, apiVip net.
 		}).Info("Failed to get master Nodes list")
 		return []Backend{}, err
 	}
-	apiVipv6 := utils.IsIPv6(apiVip)
+
+	// As it is not possible to get cluster's Machine Network directly, we are using a workaround
+	// by detecting which of the local interfaces belongs to the same subnet as requested VIP.
+	// This interface can be used to detect what was the original machine network as it contains
+	// the subnet mask that we need.
+	machineNetwork, err := utils.GetLocalCIDRByIP(apiVip.String())
```
cybertron, line 557: It makes me a little twitchy that we use machineNetwork later even if this returns an error. Maybe we just check for "" like you did elsewhere (since I think that's what machineNetwork will be in the error case)?

cybertron (commented):
I need to come back and look at this when it's not the end of the day and I'm a little more lucid, but the broad strokes look good to me.

pkg/config/node.go
```diff
@@ -291,7 +292,7 @@ func IsUpgradeStillRunning(kubeconfigPath string) (bool, error) {
 	return true, nil
 }
 
-func GetIngressConfig(kubeconfigPath string, filterIpType string) (ingressConfig IngressConfig, err error) {
+func GetIngressConfig(kubeconfigPath, apiVip, ingressVip string) (ingressConfig IngressConfig, err error) {
```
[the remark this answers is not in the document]
mkowalski: I thought about it, but the problem is that I need to have *both* vips whenever I generate a single keepalived configuration stanza. Look at the following example

API VIP - 1.1.1.101
Ingress VIP - 1.1.1.102
NodeIP - 1.1.1.50

I have a node which holds VIPs, so its list of IP addresses looks like `[1.1.1.50,1.1.1.101,1.1.1.102]`. If I passed only single VIP as a filter, now what happens...

We want to generate config for API, so we pass 1.1.1.101 to filter it out. The function will take all the IPs of the node (`[1.1.1.50,1.1.1.101,1.1.1.102]`), remove API VIP from this list so we end up with `[1.1.1.50,1.1.1.102]`. So now, the function has full authority to select `1.1.1.102` as the NodeIP to use in building the peer list. This is wrong, because `1.1.1.102` is Ingress VIP. The mistake happened because the function acted locally and had no idea there are other VIPs in this environment.

Now, if the function allows to filter out two addresses, what happens is... We want to generate config for API as previously. We pass `.101` and `.102` to filter out, saying those are both VIPs and should never make it to the NodeIP. So our list of potential addresses is very simple, `[1.1.1.50]`. This way, there is no possibility that we select a VIP as an entry in the peer list.

I agree, the naming here can be somehow confusing. I wonder if maybe I should change the function signature to have `filterIPs []string` so that you could just pass a list of IPs that should be forbidden from being selected? What do you think?

The general issue is that indeed the function `GetIngressConfig` is used to generate both API VIP and Ingress VIP configs, but can't accept only single IP for reasons described above. I guess this is just a naming confusion as "ingress" in keepalived world is different than "ingress" in OpenShift world

pkg/utils/network.go
mkowalski: LOL of course I see it now. I will clean&fix this

pkg/config/node.go
mkowalski: I see two options to solve this. Hard one and graceful one, let's see what seems wiser here

1) If we failed to get the machine network, throw an error; this feels okay now because I tend to say that for any given API VIP your node must have at least one subnet shared with the VIP subnet. On the other hand, when Control Plane across multiple subnets is introduced, this statement may not hold true any more; I think in this scenario the logic below should not be executed anyway as-is, but I can see some potential for it to fail (which may be also good because it will show us quickly what needs changing in runtimecfg to support this feature)

```
machineNetwork, err := utils.GetLocalCIDRByIP(apiVip.String())
	if err != nil {
		return []Backend{}, fmt.Errorf("Could not retrieve subnet for IP %s", apiVip.String())
	}
```

2) Inside the function doing the real calculation just check if any machine network was provided. If not, then it makes no sense to use this logic so just skip the part of the code that uses it

```
func getNodeIpForRequestedIpStack(
[...]
	if addr == "" && machineNetwork != "" {
		log.Debugf("For node %s can't find address using NodeInternalIP. Fallback to OVN annotation.", node.Name)
[...]
```

Any preference?

pkg/config/node.go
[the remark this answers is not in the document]
cybertron: Oh, I didn't think about the fact that you need to filter too. Most places we just use the VIP to determine the subnet or the ip version, but in this case you do need to ignore both. Since this was nit-picking anyway I don't think we need to change anything.

pkg/config/node.go
cybertron: I'm not sure this code will ever get called in a multiple subnet control plane (shorter name needed for this feature ;-). The internal loadbalancer will be disabled so we won't be building peer lists at all. The only question would be whether this gets called at some point by the DNS code since we're planning to keep that part. I don't _think_ it does because the only references I can find to it are in the loadbalancer monitors.

So I think I'm good with making this an error. Keepalived can't run on a node that isn't on the VIP subnet so I can't see where we would need to handle that.

pkg/utils/network.go
mkowalski: The current implementation,
```
func IsIPv6(ip net.IP) bool {
	return ip.To4() == nil
}
```
is prone to give wrong results when running against malformed IPs. I will let myself to change it, while keeping the same signature

dcbw:
What are the cases where the IP won't be part of node.Status.Addresses? Would it be if the intended IP is on a second NIC that is not the same NIC as what was given to kubelet?

mkowalski:
>What are the cases where the IP won't be part of node.Status.Addresses? Would it be if the intended IP is on a second NIC that is not the same NIC as what was given to kubelet?

Dual-stack on vSphere (or any other cloud provider, but we target only vSphere now). This is because kubelet cannot take `--cloud-provider` together with `--node-ip` when the latter contains dual-stack IPs. So for such a scenario, `--node-ip` gets IPv4 address, `--cloud-provider` gets vSphere and as a result. Node.Status.Addresses contains only IPv4. Nothing to do really with multi NIC topology at this moment.

But... Our aim was to explicitly state that till https://github.com/kubernetes/enhancements/pull/3706 is implemented, topologies where dual-stack is spawned across multi NICs are not supported because in such a scenario I am not sure if we are able to correctly unwrap&wrap traffic coming to the loadbalancer.

I somehow have a feeling that multi NIC could just work (because internally every node knows how to talk to each other and loadbalancer can use internally IPv4 even for incoming IPv6 traffic), but it's probably too cumbersome to test unless we have a customer with this specific topology

cybertron:
> I somehow have a feeling that multi NIC could just work (because internally every node knows how to talk to each other and loadbalancer can use internally IPv4 even for incoming IPv6 traffic), but it's probably too cumbersome to test unless we have a customer with this specific topology

Hmm, that's a good point. I was thinking Kubelet would use its own internal logic to select ipv6, but since it doesn't get a v6 address at all in this scenario that might not actually matter. Probably still safest to just not support that though since I'm not positive what the behavior will be.

pkg/config/node.go
```diff
@@ -306,23 +310,105 @@ func GetIngressConfig(kubeconfigPath string, filterIpType string) (ingressConfig
...
+		} else {
+			ingressConfig.Peers = append(ingressConfig.Peers, addr)
 		}
 	}
 
 	return ingressConfig, nil
 }
 
+func getNodeIpForRequestedIpStack(node v1.Node, filterIps []string, machineNetwork string) (string, error) {
+	log.SetLevel(logrus.DebugLevel)
```
cybertron, line 342: If we're forcing this on, should these log messages just be Info level?

cybertron (commented):
One minor question about the logging inline, but otherwise lgtm.

pkg/config/node.go
mkowalski: Sure, just changed. I kind of thought about a future when we want to have an env variable to control the log level to decrease verbosity, but it will be simple to then lower level from info to debug

cybertron (approved):
/lgtm

PR #219, opened 2023-01-31
OCPBUGS-3127: 5 minutes timeout for getSuitableIPs

dougsland:
Do not keep CPU in infinite loop looking for suitable IPs. It blocks ovs-configuration and also spamming syslog.


cmd/runtimecfg/node-ip.go
```diff
@@ -209,7 +209,9 @@ func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType stri
 	// Enable debug logging in utils package
 	utils.SetDebugLogLevel()
 	ipFilterFunc := utils.ValidNodeAddress
-	for {
+	const timeOutInMinutes = 5
```
mkowalski, line 212: Maybe just a NIT so feel free to ignore: I see we have a whole `const` section defined at the top of this file. Could we this timeout from here to there? This way when I open the file I immediately see all the timeouts, paths and co. in a single place

dougsland: @mkowalski sounds good to me.

cmd/runtimecfg/node-ip.go
```diff
@@ -250,10 +252,9 @@ func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType stri
 		if !retry {
 			return nil, false, fmt.Errorf("Failed to find node IP")
 		}
-
-		log.Errorf("Failed to find a suitable node IP")
 		time.Sleep(time.Second)
```
cybertron, line 255: We need to make some modifications here. Always sleeping for a second is what is spamming the logs.

cybertron (changes requested):
I don't think this is going to help with the behavior described in the bug. The issue is that we're running nodeip-configuration in a tight loop so it spams the logs with output. Timing out after 5 minutes is still going to result in hundreds of loops. On top of that, when this fails the systemd service is going to restart it anyway so the timeout won't be effective.

Also, I don't think we want a timeout. nodeip-configuration _must_ complete before other services can start. What we need is an exponential backoff of the sleep time in between iterations so if there's a perpetual failure we'll only run once every 5 minutes or so.

As an example, here's what the loop looks like right now in a perma-fail env:
run ip detection
sleep 1
run ip detection
sleep 1
run ip detection
sleep 1
...repeat indefinitely...

What we want is something like this:
run ip detection
sleep 1
run ip detection
sleep 2
run ip detection
sleep 4
...
run ip detection
sleep 300
run ip detection
...repeat indefinitely...

dougsland:
@cybertron  Looks like everything is working nice, please double check if your devscript takes the normal time to build and run:

```
vel=debug msg=Cluster is initialized
level=info msg=Checking to see if there is a route at openshift-console/console...
level=debug msg=Route found in openshift-console namespace: console
level=debug msg=OpenShift console route is admitted
level=info msg=Install complete!
level=info msg=To access the cluster as the system:admin user when using 'oc', run 'export KUBECONFIG=/home/douglas/dev-scripts/ocp/ostest/auth/kubeconfig'
level=info msg=Access the OpenShift web-console here: https://console-openshift-console.apps.ostest.test.metalkube.org
level=debug msg=Time elapsed per stage:
level=debug msg=         bootstrap: 16s
level=debug msg=           masters: 18m37s
level=debug msg=Bootstrap Complete: 21m27s
level=debug msg= Bootstrap Destroy: 10s
level=debug msg= Cluster Operators: 39m54s
level=info msg=Time elapsed: 1h20m38s
```

cmd/runtimecfg/node-ip.go
```diff
@@ -206,10 +208,15 @@ func checkAddressUsable(chosen []net.IP) (err error) {
 }
 
 func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType string) (chosen []net.IP, matchesVips bool, err error) {
+	// timerLoop will hold a time in Seconds to be used with time.Sleep() before going
+	// for the next loop interation.
+	timerLoop := 0
```
cybertron, line 213: We may not want to start this at 0. Even on the first retry we want there to be _some_ delay.

cmd/runtimecfg/node-ip.go
```diff
@@ -252,7 +259,10 @@ func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType stri
 		}
 
 		log.Errorf("Failed to find a suitable node IP")
-		time.Sleep(time.Second)
+		time.Sleep(time.Second * time.Duration(timerLoop))
+		if timerLoop == maxSecondsToSuitableIPsLoop {
```
mkowalski, line 263: This part is fragile, let me show you why - we start from `1` and then we add `addSecondsToSuitableIPsLoop` every time we loop. So in our case we start with `1`, add `2` so after the 1st iteration we have `3`. Then we again add `2` so we have `5`... And so on, till we reach `maxSecondsToSuitableIPsLoop` which now is `300`.

But this assumes that this repetitive addition will get you to exactly `maxSecondsToSuitableIPsLoop` at some point. So what happens if someone in the future changes `addSecondsToSuitableIPsLoop` from `2` to let's say `7`. This way you will never have `timerLoop == maxSecondsToSuitableIPsLoop` returning `true`, because of how `1+7+7+7+7...` works

So tldr it would be safer to have `timerLoop >  maxSecondsToSuitableIPsLoop`

cmd/runtimecfg/node-ip.go
```diff
@@ -252,7 +259,10 @@ func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType stri
...
+			timerLoop = 1
```
cybertron, line 264: We don't want to reset to 1 here, we want to cap it at 300. Once we hit a 5 minute delay we want to keep delaying 5 minutes until it succeeds. That's the key part of reducing the log spam.

cybertron (changes requested):

cmd/runtimecfg/node-ip.go
dougsland: that make sense.

cmd/runtimecfg/node-ip.go
dougsland: gotcha

cmd/runtimecfg/node-ip.go
```diff
@@ -206,10 +208,15 @@ func checkAddressUsable(chosen []net.IP) (err error) {
...
 func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType string) (chosen []net.IP, matchesVips bool, err error) {
+	// timerLoop will hold a time in Seconds to be used with time.Sleep() before going
+	// for the next loop interation.
+	timerLoop := 1
+
 	// Enable debug logging in utils package
 	utils.SetDebugLogLevel()
 	ipFilterFunc := utils.ValidNodeAddress
 	for {
+		timerLoop = timerLoop + addSecondsToSuitableIPsLoop
```
cybertron, line 219: Oh, I didn't catch this the first time, but is there a reason you're adding instead of multiplying here? That will make it take 150 iterations to get to a 5 minute delay

cmd/runtimecfg/node-ip.go
```diff
@@ -252,7 +259,13 @@ func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType stri
 		}
 
 		log.Errorf("Failed to find a suitable node IP")
-		time.Sleep(time.Second)
+		if timerLoop > maxSecondsToSuitableIPsLoop {
```
cybertron, line 262: Probably not a realistic problem, but I would cap timerLoop to maxSeconds... instead of changing the Sleep call. No chance of timerLoop overflowing its type that way.

cybertron (commented):
One other note on the commit message: "It blocks ovs-configuration" <- this is intentional. ovs-configuration can't happen until we have ips.

cmd/runtimecfg/node-ip.go
dougsland: good one.

cmd/runtimecfg/node-ip.go
dougsland: Please see if the next patch address the request.

mkowalski:
/test e2e-metal-ipi

Doesn't seem like the failures relate to this change, but for the sake of completeness retesting it

cybertron:
/retest

Seeing a lot of red in the metal-ipi job history, which makes me suspicious there is an unrelated problem with the job. Let's see if it's working yet.

Given that it's not required and the ipv6 job is green, I'm not inclined to spend too much time retesting. If it doesn't pass this time and doesn't look related to the patch I may just merge it.

PR #220, opened 2023-02-01
OPNET-214: node-ip: adjust logic for user-managed-lb

EmilienM:
Introducing a new option (false by default): `--user-managed-lb`
If this option is set, we'll not mark a worker node as a remote-worker
if a chosen IP doesn't match the VIPs.

Since the VIPs are managed externally, we don't need to care about it.

### DIFF
--- a/cmd/runtimecfg/node-ip.go
+++ b/cmd/runtimecfg/node-ip.go
@@ -27,7 +27,7 @@ const (
 	ovn                      = "OVNKubernetes"
 )
 
-var retry, preferIPv6 bool
+var retry, preferIPv6, userManagedLB bool
 var networkType string
 var nodeIPCmd = &cobra.Command{
 	Use:                   "node-ip",
@@ -69,6 +69,7 @@ func init() {
 	nodeIPCmd.PersistentFlags().BoolVarP(&retry, "retry-on-failure", "r", false, "Keep retrying until it finds a suitable IP address. System errors will still abort")
 	nodeIPCmd.PersistentFlags().BoolVarP(&preferIPv6, "prefer-ipv6", "6", false, "Prefer IPv6 addresses to IPv4")
 	nodeIPCmd.PersistentFlags().StringVarP(&networkType, "network-type", "n", ovn, "CNI network type")
+	nodeIPShowCmd.Flags().BoolVarP(&userManagedLB, "user-managed-lb", "l", false, "User managed load balancer")
 	rootCmd.AddCommand(nodeIPCmd)
 }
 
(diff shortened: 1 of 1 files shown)


cmd/runtimecfg/node-ip.go
```diff
@@ -113,10 +114,11 @@ func set(cmd *cobra.Command, args []string) error {
...
 	remoteWorker := false
-	// if chosen ip doesn't match vips need create a file that
-	// this file will be used by keepalived container to verify if it should run or not
-	// We want to disable keepalive in case this host is remote worker
-	if len(vips) > 0 && !matchesVips {
+	// if chosen ip doesn't match vips, we need create a file that
+	// will be used by keepalived container to verify if it should run or not
+	// We want to disable keepalived in case this host is remote worker
+	// This is skipped in case of user managed load balancer as the VIPs are not managed by keepalived.
+	if len(vips) > 0 && !matchesVips && !userManagedLB {
```
tsorya, line 121: This is not the right place, this one ensures that keepalived will sleep forever and not remote-worker label

mandre, line 121: Does it mean that remote worker node becomes incompatible with user-managed LB? Are we fine with it?

mdbooth:
The problem is it's obtuse. The feature seems to be about remote workers, but the flag is about load balancers. As a user, that means I need to understand that some chain of implementation details means that one affects the other. And what happens in the future when we change the implementation and they don't?

Please add a flag which has a user-comprehensible intent.

cmd/runtimecfg/node-ip.go
EmilienM: That doesn't make me happy but we haven't come up with anything better so far.
@tsorya @cybertron could you think of another way to detect if a node is a remote-worker?

EmilienM: I'll be nice if we wouldn't block remote workers to work with an external LB.

cybertron: This doesn't block remote workers with external LB, does it? Keepalived won't be deployed on those nodes in either case. If you're using external LB then we don't need the remote worker logic because everything is essentially remote.

tsorya: Yep, agree with @cybertron. With external LB all the workers are actually remote one so it doesn't really matters

EmilienM: I'm talking about the labelling, which won't happen if remote-worker + external-LB (in this code).

EmilienM: we discussed this offline and I think this is a viable option now. Since we don't need labelling when an external LB is in play anyway.

PR #223, opened 2023-02-03
OPNET-210: Support preferIPv6 for IPI too

cybertron:
We added support for the preferIPv6 option in node-ip selection for UPI, but since we always preferred IPv4 in IPI except in v6-only deployments it wasn't wired in for the IPI path. This patch fixes that so we can do v6-primary dual stack deployments and select the appropriate node ip.


dougsland (commented):
/lgtm

PR #224, opened 2023-02-07
OPNET-133: Enabling remote worker feature only for baremetal platform

tsorya:


cmd/runtimecfg/node-ip.go
```diff
@@ -269,3 +270,8 @@ func parseIPs(args []string) ([]net.IP, error) {
 	}
 	return ips, nil
 }
+
+// currently we allow setting remote worker only in case of baremetal platform without external lb
+func setRemoteWorker(vips []net.IP, matchesVips bool) bool {
```
mkowalski, line 275: Can we rename from `set` to `is`? This function as I see it is only answering a question and is not setting any state anywhere

cybertron:
/lgtm

I don't see anything ip version specific here so as long as the required ipv6 job passes we should be good.

PR #228, opened 2023-03-27
node.go: remove no required check

dougsland:
There is no err return in IsTheSameConfig()


pkg/config/node.go
```diff
@@ -281,11 +281,8 @@ func IsUpgradeStillRunning(kubeconfigPath string) (bool, error) {
 	// Go to all node types identified in GetNodes()
 	for nodeRole := range nodes {
 		nodesConfigs := IsTheSameConfig(nodes[nodeRole])
-		if err != nil {
```
mkowalski, line 284: Seems to me like that was dead code anyway, am I right? Because if this err happened, it would have been thrown in L277 just above? I will LGTM tomorrow morning because it seems like pretty straightforward thing

PR #230, opened 2023-03-29
Fixup isUpgradeStillRunning logic

fherbert:
Only return false from isUpgradeStillRunning when ALL node groups have same config, not just one of them, also, remove redundant error check.


dougsland:
Thanks @fherbert !

PR #231, opened 2023-03-30
OCPBUGS-11138: fix isUpgradeStillRunning()

dougsland:
- node.go: remove no required check
- Only return false from isUpgradeStillRunning when ALL node groups have same config, not just one of them


cybertron:
/retest

I don't see how this could have broken anything, but I'd just as soon see all the jobs pass, even the non-required ones.

dougsland:
@amalykhi @reihl could you please review this patch and add cherry-pick-approve?  Thanks!

reihl (approved):

PR #233, opened 2023-04-14
OCPBUGS-11297: fix isUpgradeStillRunning()

dougsland:
- node.go: remove no required check
- Only return false from isUpgradeStillRunning when ALL node groups have same config, not just one of them


dougsland:
@reihl  could you please cherry approve?

reihl (approved):

PR #237, opened 2023-04-25
OCPBUGS-12729: Make nested dual stack VIP configs respect EnableUnicast

cybertron:
Currently the nested configs that were used to enable dual stack VIPs do not respect the EnableUnicast setting correctly, which in the case of EUS upgrades that skip 4.11 can cause problems with the unicast migration being triggered at an incorrect time.

This patch just propagates the EnableUnicast setting into the nested configs so it will apply correctly in dual stack VIP environments.


cybertron:
/hold

This isn't working right yet.

PR #238, opened 2023-04-25
DO NOT MERGE: Test PR to build release image with enable unicast fix

cybertron:
Currently the nested configs that were used to enable dual stack VIPs do not respect the EnableUnicast setting correctly, which in the case of EUS upgrades that skip 4.11 can cause problems with the unicast migration being triggered at an incorrect time.

This patch just propagates the EnableUnicast setting into the nested configs so it will apply correctly in dual stack VIP environments.

(cherry picked from commit 64751219c12da84d36227e6ab87b26b5ac2a4bc7)


cybertron:
Confirmed that this works. Will backport with the bot though.

PR #257, opened 2023-06-08
OCPBUGS-14403: Use machine-config state instead of comparing roles

cybertron:
It turns out that the mapping of roles to machine-configs is more complex than we realized when we first implemented the unicast migration and it's causing bugs in more complex environments with nodes in multiple roles. The machineconfiguration.openshift.io/state annotation directly provides us the upgrade status of the node anyway, so there's no reason to try to derive it ourselves. I'm unsure why we didn't use that in the first place, but I don't see any reason not to.

Note that it was also proposed to just use the kubelet version check we previously added to handle EUS upgrades. However, in my testing that doesn't always align perfectly with the node's state so I'm concerned that nodes may schedule the migration incorrectly if we rely only on that. Using both kubelet version and state is still much simpler than what we were doing before and should cover all of the use cases I'm aware of.

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -197,24 +197,14 @@ func GetVRRPConfig(apiVip, ingressVip net.IP) (vipIface net.Interface, nonVipAdd
 	return getInterfaceAndNonVIPAddr(vips)
 }
 
-// GetNodes will collect all nodes which contain the annotation node-role.kubernetes.io/${VAR} annotation
-// and return a map with the data.
-//
-// i.e:
-//
-//	node-role.kubernetes.io/master
-//	node-role.kubernetes.io/worker
-//	node-role.kubernetes.io/infra
-//	etc.
+// GetNodes will return a list of all nodes in the cluster
 //
 // Args:
 //   - kubeconfigPath as string
 //
 // Returns:
-//   - map[string][]v1.Node or error
-func GetNodes(kubeconfigPath string) (map[string][]v1.Node, error) {
-	nodeCluster := make(map[string][]v1.Node)
-
+//   - v1.NodeList or error
+func GetNodes(kubeconfigPath string) (*v1.NodeList, error) {
 	config, err := utils.GetClientConfig("", kubeconfigPath)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


cybertron:
Leaving in draft so I can test this locally and verify it does what we need.

cybertron:
I discovered another problem with the upgrade complete detection logic. A second commit has been added that fixes it. Leaving in draft because I still need to test the fix e2e, but I've confirmed the behavior in isolation so I expect this should work.

cybertron:
I have confirmed that this is now doing the right thing on upgrade. We should be good to merge it.
/cherry-pick release-4.13

pkg/config/node.go
```diff
@@ -229,74 +219,40 @@ func GetNodes(kubeconfigPath string) (map[string][]v1.Node, error) {
...
 		// intermediate state where all of the nodes are "updated" as far as
 		// MCO is concerned, but are actually on different versions of OCP.
 		// In those cases, we do not consider the upgrade complete because not
 		// all nodes are ready for migration.
 		if kubeletVersion == "" {
-			kubeletVersion = nodes[nodeRole][0].Status.NodeInfo.KubeletVersion
+			kubeletVersion = node.Status.NodeInfo.KubeletVersion
 		}
-		if kubeletVersion != nodes[nodeRole][0].Status.NodeInfo.KubeletVersion {
+		if kubeletVersion != node.Status.NodeInfo.KubeletVersion {
```
mkowalski, line 252: It may be not relevant at this moment, but I wonder if for heterogeneous clusters the `KubeletVersion` string is always the same. I think the answer is yes because I see `kubeletVersion: v1.25.4+77bec7a` so it contains version and git hash which would mean we have a platform for which kubelet was built based on a different commit, but nevertheless I want to raise this point because it may be useful in the future if we see something strange happening

mkowalski:
/lgtm

It simply looks good. I'm with you wondering why not relying on this in the first place and hope this time it fixes the issue for good

pkg/config/node.go
cybertron: It's a good point and in a perfect world I would redesign all of this so we're not trying to guess what version all of the nodes are on, but at this point I don't think we can make drastic changes in 4.11. Of all the fields we have access to I think kubelet version is the least likely to vary within a cluster.

cybertron:
> It simply looks good. I'm with you wondering why not relying on this in the first place and hope this time it fixes the issue for good

Just for historical reference in case anyone needs to come back to this, I suspect the reason we didn't use this field initially is that the original implementation was done for baremetal only way back in 4.7. I bet that field didn't exist in that release and was only added since. Now that we only care about migrations in 4.11 we can use the field.

pkg/config/node.go
palonsoro: I have read this now. Not sure if this comment will be of any help this late @cybertron , but just in case:
- In RHCOS, kubelet version is dictated by the RPM on the RHCOS version, so it is the same on all the RHCOS nodes.
- On the other hand, in RHEL nodes, it gets whatever kubelet version it can from the yum repos, so the kubelet version for RHEL nodes might be different.
- However, even if there is mismatch between RHEL nodes and RHCOS nodes, it would be only in the `.z` part, i.e. you may have a 1.24.a node and a 1.24.b node if one is RHCOS and the other one is RHEL, but you would never have a 1.24.z node and a 1.25.z node on a cluster unless it is in the middle of an upgrade (including an EUS upgrade) or somebody did something unsupported.

palonsoro:
> Just for historical reference in case anyone needs to come back to this, I suspect the reason we didn't use this field initially is that the original implementation was done for baremetal only way back in 4.7. I bet that field didn't exist in that release and was only added since. Now that we only care about migrations in 4.11 we can use the field.

Just for the record and in case it is required: The machine-config annotations, including the state, have existed in 4.7 and before, so I don't believe that was the reason.

The only reason for the older more complex logic that comes to my mind would be to avoid a scenario where all nodes are "Done" because a subset is like `oldDesired==oldCurrent` and the rest are already `newDesired==newCurrent`, i.e. avoid problems in the middle of the upgrade when some node completed but machine-config-controller has not instructed the next one to upgrade by setting the desired label.

However, at that time the kubelet version check was not there. If the kubelet versions vary (which is most likely), there would be no problem. And I am not even sure of the real chances of hitting the scenario I describe above, given the quite unlucky timings we would need.

PR #258, opened 2023-06-09
OCPBUGS-12739: Fix bug with IP conversion using net.IP

mkowalski:
Fixing a bug where instead of `net.ParseIP()` we called `net.IP()`. This was causing a malformed IP address to be passed around which ultimately was causing us to always think we are in IPv4 setup.

Contributes-to: OCPBUGS-12739


dougsland:
LGTM

PR #262, opened 2023-06-22
[release-4.12] OCPBUGS-15315: Use machine-config state instead of comparing roles

cybertron:
This is a manual cherry-pick of #261 . It did not backport cleanly due to changes in how the client for GetNodes is created. Otherwise, all changes applied cleanly.


zhaozhanqi:
@cybertron @mkowalski  need you label `backport-risk-assessed`

PR #269, opened 2023-08-24
OCPBUGS-18096: Don't render config with incomplete unicast peer list

cybertron:
If there is an error retrieving either the ingress or api peer list, we should not continue with config rendering. If there are three identical failures in a row it will cause us to render a config with no unicast peers, which can cause a split-brain with other nodes that are still using unicast.

This adds a return of the error from updateUnicastConfig and handles it in the calling code. In the case of the migration code, we loop until the list is retrieved successfully since we can't proceed without peers. In the case of a normal config update we just skip that iteration and retry on the next time interval.

(cherry picked from commit 3b4938591a1ecd4cdae89fb7eb97d155d2eb15bd)

This did not cherry pick cleanly because of the dual stack VIP work in 4.12.

### DIFF
--- a/pkg/monitor/dynkeepalived.go
+++ b/pkg/monitor/dynkeepalived.go
@@ -58,21 +58,24 @@ func getActualMode(cfgPath string) (error, bool) {
 	return nil, enableUnicast
 }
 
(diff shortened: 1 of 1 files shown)


mkowalski (approved):
I am surprised how we carried this unused `appliedConfig` var here, but I guess it is what it is

/lgtm

PR #270, opened 2023-09-01
OCPBUGS-18257: Move haproxy firewall rule check earlier in loop

cybertron:
When we stopped rendering config if we were unable to update the peer list[0] we also stopped checking the haproxy firewall rule status if there was a problem contacting the API. What this means is that we may not correctly update the priority of a node with no access to the API, causing the VIP to stay there when it should migrate to a different node.

I think this will mostly happen during initial deployment when the apiserver may become briefly unavailable. If this happens haproxy will remove the firewall rule so we fall back to direct un-loadbalanced API access, but if there isn't an apiserver instance running on the node yet that won't work either. There's no way for the healthchecks to recover from this situation.

To fix this I just moved the firewall rule check to be first in the loop flow so it will get run even if we bail out early due to a failure in contacting the API.

0: 18bc0e360f6430fd1ec52db83b77841c30c5e8fa

### DIFF
--- a/pkg/monitor/dynkeepalived.go
+++ b/pkg/monitor/dynkeepalived.go
@@ -407,6 +407,34 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
 			appliedConfig = curConfig
 
 		default:
+			// Signal to keepalived whether the haproxy firewall rule is in place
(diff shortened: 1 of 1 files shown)


mkowalski:
/label acknowledge-critical-fixes-only

Merging this will help pushing https://issues.redhat.com/browse/OCPBUGS-18530 forward which has some sense of urgency

PR #276, opened 2023-09-25
OCPBUGS-19492: Increase timeout for bootstrap kubeapi

mkowalski:
We are increasing timeout for waiting for boostrap keepalived from the current 30 seconds to 30 minutes. This is because we observed environments where 30 seconds is not enough for the kube-api to start yet it does start eventually after much longer time.

We don't understand why kube-api is so slow on the bootstrap node in those cases but as this timeout has no any deeper meaning it is safe to just increase it to extremely big value.

Contributes-to: OCPBUGS-19492


mkowalski:
/hold
/cherry-pick release-4.14

Pre-merge testing should be done as we aim to backport to OCP 4.12 (so should be clear that this really fixes the issue)

cybertron:
/lgtm
/retest-required

Once pre-merge testing is complete, anyone can feel free to remove the hold and let this merge.

mkowalski:
/unhold

We don't need anything to get it to master

PR #286, opened 2023-11-05
CORS-2818: Adding ability to render Cloud LB IPs

sadasu:
Add the ability to render and monitor Cloud LB IPs. This capability is used only when `userProvisionedDNS` is enabled on some (AWS, GCP and Azure) cloud platforms via install-config.
    
On some cloud platforms (AWS, Azure, GCP), the user cannot use the cloud's default DNS solution. They are expected to use their own custom DNS solution that is external to the cluster. OpenShift is not allowed to configure this DNS solution. In this scenario, these same customers want to continue using the cloud provided Load Balancers (LBs).
    
OpenShift is expected to continue configuing the cloud LBs for API, API-Int and Ingress access. Since the LB information is not available, before cluster installation, the user cannot configure their custom DNS solution before cluster installation.
    
So, to support this mode, OpenShift needs to start its in-cluster CoreDNS based DNS solution for API, API-Int and Ingress resolution so that cluster installation is successful. The customer is expected to configure their DNS solution post-install.
    
At this the time, the cloud's API and API-Int LBs are configured by the Installer and its value is not expected to change during the life of the cluster. The Ingress operator continues to handle Ingress LB configuration.
    
Based on enhancement: https://github.com/openshift/enhancements/pull/1468

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -52,6 +52,11 @@ type Cluster struct {
 	VIPNetmask             int
 	MasterAmount           int64
 	NodeAddresses          []NodeAddress
+	APILBIPs               []string
+	APIIntLBIPs            []string
+	IngressLBIPs           []string
+	CloudLBRecordType      string
+	CloudLBEmptyType       string
 }
 
 type Backend struct {
@@ -82,6 +87,12 @@ type Node struct {
 	IngressConfig IngressConfig
 	EnableUnicast bool
 	Configs       *[]Node
+}
+
+type ClusterLBConfig struct {
+	ApiLBIPs     []net.IP
+	ApiIntLBIPs  []net.IP
+	IngressLBIPs []net.IP
 }
 
 func getDNSUpstreams(resolvConfPath string) (upstreams []string, err error) {
@@ -417,7 +428,13 @@ func getNodeIpForRequestedIpStack(node v1.Node, filterIps []string, machineNetwo
 // apiPort: The port on which the k8s api listens. Should be 6443.
 // lbPort: The port on which haproxy listens.
 // statPort: The port on which the haproxy stats endpoint listens.
-func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVips, ingressVips []net.IP, apiPort, lbPort, statPort uint16) (node Node, err error) {
+// clusterLBConfig: A struct containing IPs for API, API-Int and Ingress LBs
+func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVips, ingressVips []net.IP, apiPort, lbPort, statPort uint16, clusterLBConfig ClusterLBConfig) (node Node, err error) {
+	if len(clusterLBConfig.ApiIntLBIPs) > 0 {
+		// Cloud Platforms with cloud LBs but no Cloud DNS
+		return getNodeConfigWithCloudLBIPs(kubeconfigPath, clusterConfigPath, resolvConfPath, clusterLBConfig)
+	}
+	// On-prem platforms
 	vipCount := 0
 	if len(apiVips) > len(ingressVips) {
 		vipCount = len(apiVips)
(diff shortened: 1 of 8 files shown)


mkowalski (changes requested):

pkg/config/node.go
```diff
@@ -437,7 +443,7 @@ func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVips
 		} else {
 			ingressVip = nil
 		}
-		newNode, err := getNodeConfig(kubeconfigPath, clusterConfigPath, resolvConfPath, apiVip, ingressVip, apiPort, lbPort, statPort)
+		newNode, err := getNodeConfig(kubeconfigPath, clusterConfigPath, resolvConfPath, apiVip, ingressVip, apiPort, lbPort, statPort, apiLBIPs[0], apiIntLBIPs[0])
```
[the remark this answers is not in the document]
sadasu: Adding unit tests. To answer the other question, apiLBIPs[0], apiIntLBIPs[0] would be "" by default. It will have actual values only when the `userProvisionedDNS` feature is enabled.

sadasu:
/retest-required

Retesting after https://github.com/openshift/installer/pull/7671 merged.

pkg/config/node.go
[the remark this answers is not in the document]
mkowalski: Sorry if I miss something obvious but I don't see that... You call `GetConfig` by passing `[]net.IP{}` as `apiLBIPs` and then you reference `apiLBIPs[0]`. This does not work, try running that code

```
a := []net.IP{}
fmt.Println(a[0])
```

and the ourput is not `""` but a panic `runtime error: index out of range [0] with length 0`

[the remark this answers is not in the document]
sadasu: Thanks @mkowalski ! I needed to update the logic in that part of the code too.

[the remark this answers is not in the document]
mkowalski: I would really recommend starting from unit tests and then proceed with the implementation... What I can see in the current code is that it is still panicking because `[]net.IP{}` and `nil` are different entities (i.e. empty slice and `nil` is not the same in golang).

Look at the following code

```
a := []net.IP{}
if a == nil {
	fmt.Println("nil")
} else {
	fmt.Println(a[0])
}
```

or run it via playground, e.g. https://go.dev/play/p/UbrxBrUstTw

mkowalski:
/hold

Some fundamental flaws in error handling, lets work on them first

sadasu:
@mkowalski I think the PR is ready for /hold cancel

@cybertron , @mkowalski , @davegord Could you PTAL?

mkowalski:
Sorry, I still don't see a single unit test here. Can you please add tests covering the code added by this PR? I just want to make sure it will be future-proof as you are implementing a feature not in direct scope of our team but in a repository mainly managed by us. It means that without test coverage this will be a zombie code very soon

mkowalski:
/test e2e-metal-ipi-ovn-ipv6

This job has to pass

pkg/config/cloud_lb_config_test.go
```diff
@@ -0,0 +1,55 @@
+package config
```
mkowalski, line 1: NIT: all this content rather goes to `node_test.go` file as it's only 2 test cases, so no need to have a dedicated file for them

pkg/config/cloud_lb_config_test.go
```diff
@@ -0,0 +1,55 @@
...
+	testIngressTwoIPv4    = net.ParseIP("10.10.10.40")
+	testClusterLBConfig   = ClusterLBConfig{
+		ApiLBIPs:     []net.IP{testApiLBIPv4},
+		ApiIntLBIPs:  []net.IP{testApiIntLBIPv4},
+		IngressLBIPs: []net.IP{testIngressOneIPv4, testIngressTwoIPv4}}
+	expectedApiLBIPv4      = "192.168.0.111"
+	expectedApiIntLBIPv4   = "10.10.10.20"
+	expectedIngressOneIPv4 = "192.168.20.140"
+	expectedIngressTwoIPv4 = "10.10.10.40"
+	updateNode             = Node{}
```
mkowalski, line 26: No need for those 2 to be global as they do not carry any state that needs to be shared

pkg/config/node.go
```diff
@@ -417,34 +428,42 @@ func getNodeIpForRequestedIpStack(node v1.Node, filterIps []string, machineNetwo
...
-		if i < len(apiVips) {
-			apiVip = apiVips[i]
-		} else {
-			apiVip = nil
-		}
-		if i < len(ingressVips) {
-			ingressVip = ingressVips[i]
+// clusterLBConfig: A struct containing IPs for API, API-Int and Ingress LBs
+func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVips, ingressVips []net.IP, apiPort, lbPort, statPort uint16, clusterLBConfig ClusterLBConfig) (node Node, err error) {
+	switch {
```
mkowalski, line 433: This switch-case adds a level of indentation to the whole function where it's not really needed (i.e. right now the whole body of this function minus single line is written inside an `else`/`default` statement).

As you are immediately returning after your check you can simplify it so instead of

```
switch {
  case XXX:
    return blablabla
  default:
    lots of code
```

you can write

```
if XXX:
  return blablabla
lots of code
```

(note I am skipping `else` completely in my revised version).

pkg/config/node.go
```diff
@@ -694,3 +713,80 @@ func PopulateNodeAddresses(kubeconfigPath string, node *Node) {
 		}
 	}
 }
+
+func getConfigWithCloudLBIPs(kubeconfigPath, clusterConfigPath, resolvConfPath string, clusterLBConfig ClusterLBConfig) (node Node, err error) {
+	var apiLBIP, apiIntLBIP, ingressIP net.IP
+	nodes := []Node{}
+	ipCount := 0
+	if len(clusterLBConfig.ApiIntLBIPs) > len(clusterLBConfig.IngressLBIPs) {
```
mkowalski, line 721: When is it possible that we have different number of IPs? This is a honest question, in the IPI loadbalancer we do there is always a requirement that you have the same number of IPs for API and for Ingress

pkg/config/node.go
```diff
@@ -694,3 +713,80 @@ func PopulateNodeAddresses(kubeconfigPath string, node *Node) {
...
+		ipCount = len(clusterLBConfig.ApiIntLBIPs)
+	} else {
+		ipCount = len(clusterLBConfig.IngressLBIPs)
+	}
+	for i := 0; i < ipCount; i++ {
```
mkowalski, line 726: As discussed on Slack, rather stick with `for _, item := range items` unless you really need to iterate inside the loop in the C++ way

pkg/config/node.go
```diff
@@ -694,3 +713,80 @@ func PopulateNodeAddresses(kubeconfigPath string, node *Node) {
...
+		if i < len(clusterLBConfig.ApiIntLBIPs) {
+			apiIntLBIP = clusterLBConfig.ApiIntLBIPs[i]
+		} else {
+			apiIntLBIP = nil
```
mkowalski, line 730: Because in L718 you wrote `var apiIntLBIP net.ip` you do not need to assign it to `nil` explicitly as it's the value you get from start

pkg/config/node.go
```diff
@@ -694,3 +713,80 @@ func PopulateNodeAddresses(kubeconfigPath string, node *Node) {
...
+		// LBs so apiLBIPs could be empty.
+		if len(clusterLBConfig.ApiLBIPs) != 0 && i < len(clusterLBConfig.ApiLBIPs) {
+			apiLBIP = clusterLBConfig.ApiLBIPs[i]
+		} else {
+			apiLBIP = nil
+		}
+		if len(clusterLBConfig.IngressLBIPs) != 0 && i < len(clusterLBConfig.IngressLBIPs) {
+			ingressIP = clusterLBConfig.IngressLBIPs[i]
+		} else {
+			ingressIP = nil
```
mkowalski, line 743: As above, no need for an explicit nil-assignment

pkg/config/node.go
```diff
@@ -417,34 +428,42 @@ func getNodeIpForRequestedIpStack(node v1.Node, filterIps []string, machineNetwo
...
-			apiVip = nil
-		}
-		if i < len(ingressVips) {
-			ingressVip = ingressVips[i]
+// clusterLBConfig: A struct containing IPs for API, API-Int and Ingress LBs
+func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVips, ingressVips []net.IP, apiPort, lbPort, statPort uint16, clusterLBConfig ClusterLBConfig) (node Node, err error) {
+	switch {
+	case len(clusterLBConfig.ApiIntLBIPs) > 0:
+		// Cloud Platforms with cloud LBs but no Cloud DNS
+		return getConfigWithCloudLBIPs(kubeconfigPath, clusterConfigPath, resolvConfPath, clusterLBConfig)
```
mkowalski, line 436: NIT: rename from `getConfigWithCloudLBIPs` to `getNodeConfigWith...` because we already have function `getNodeConfig` and yours is just wrapping it with some additional commands

pkg/monitor/corednsmonitor.go
```diff
@@ -42,10 +42,15 @@ func CorednsWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath strin
...
 			}
-			newConfig, err := config.GetConfig(kubeconfigPath, clusterConfigPath, resolvConfFilepath, apiVips, ingressVips, 0, 0, 0)
+			clusterLBConfig := config.ClusterLBConfig{ApiLBIPs: apiLBIPs, ApiIntLBIPs: apiIntLBIPs, IngressLBIPs: ingressLBIPs}
+			newConfig, err := config.GetConfig(kubeconfigPath, clusterConfigPath, resolvConfFilepath, apiVips, ingressVips, 0, 0, 0, clusterLBConfig)
 			if err != nil {
 				return err
 			}
+			// Populate cloud LB IP addresses for platforms where the cloud LBs
+			// have already been configured
+			newConfig, _ = config.PopulateCloudLBIPAddresses(clusterLBConfig, newConfig)
```
mkowalski, line 52: Let's not ignore errors here (anywhere, in general)

mkowalski:
Hey, reviewed the last revision. The overall design seems okay, there is a couple of language-specific nits + more test cases needed (mainly for error handling and edge cases of passing empty values)

pkg/config/node.go
sadasu: In the cloud IPI case with BYO DNS, there are internal and external Ingress LBs. That is one way to have multiple entries for Ingress. For private clusters, we don't expect to have any IPs for API LB IPs, only API-Int and Ingress LBs are expected to be running. This is a way to future-proof our implementation for all combinations of LB IPs.

pkg/config/node.go
sadasu: apiIntLBIP and other LB IP values are being set within a loop starting at L726 so its value could have been !nil in the previous iteration.

mkowalski: Yeah, I see it now... In that case this deserves at least a comment in Simple English directly in the code or some kind of a simplification...

As I read this whole loop now (L726-L730) is that when you finish the loop, you want `apiIntLBIP` to have a value of the last element of the `clusterLBConfig.ApiIntLBIPs` slice. Is that correct? If yes, in pseudocode that would be

```
apiIntLBIP = get_last_element(clusterLBConfig.ApiIntLBIPs)
```

Do I get the desired behaviour right? If yes, I'd rather do something like

```
apiIntLBIP = clusterLBConfig.ApiIntLBIPs[len(clusterLBConfig.ApiIntLBIPs)-1]
```

instead of iteration in which every time we assign a value only to ignore it afterwards and use the last assignment.

If my reasoning is wrong and your desired behaviour is different then ignore what I wrote, but that means a comment is more than desired

sadasu: I was trying to follow the precedence set by https://github.com/openshift/baremetal-runtimecfg/blob/master/pkg/config/node.go#L421-L447. There are definitely other ways to implement this too. 
Adding a comment for clarity.

mkowalski:
/hold cancel
/lgtm

/cc @cybertron 
The code looks sane, please look if the overall design is okay

PR #287, opened 2023-11-20
OCPBUGS-23432: Use shorter IP label for keepalived VIP

mkowalski:
This PR modifies the label used for an IP address assigned by keepalived to not include the name of the interface. This is because the maximum allowed lenght of the label is 15 characters, causing stuff to fail when interface itself has a name longer than 12 characters.

With this change the label will always be a simple string of length 3 and will not be susceptible to overflow for long NIC names.

Contributes-to: OCPBUGS-23432

### DIFF
--- a/test/data/keepalived.conf.tmpl
+++ b/test/data/keepalived.conf.tmpl
@@ -21,7 +21,7 @@ vrrp_instance {{.Cluster.Name}}_API {
         auth_pass {{.Cluster.Name}}_api_vip
     }
     virtual_ipaddress {
-        {{.Cluster.APIVIP}}/{{.Cluster.VIPNetmask}} label {{.VRRPInterface}}:vip
+        {{.Cluster.APIVIP}}/{{.Cluster.VIPNetmask}} label vip
     }
     track_script {
         chk_ocp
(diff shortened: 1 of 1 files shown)


ricky-rav:
/lgtm
Thanks, Mat!

mkowalski:
/approve

Because I can :)

mkowalski:
/retest

Test infra is choking

test/data/keepalived.conf.tmpl
```diff
@@ -39,7 +39,7 @@ vrrp_instance {{.Cluster.Name}}_INGRESS {
         auth_pass {{.Cluster.Name}}_ingress_vip
     }
     virtual_ipaddress {
-        {{.Cluster.IngressVIP}}/{{.Cluster.VIPNetmask}} label {{.VRRPInterface}}:vip
+        {{.Cluster.IngressVIP}}/{{.Cluster.VIPNetmask}} vip
```
cybertron, line 42: Did you mean to drop "label" from this line?

mkowalski: No, of course no. Luckily this code here is just a test and not runtime

cybertron (approved):
/lgtm

PR #292, opened 2024-01-10
OCPBUGS-26765: Add .snyk file to ignore vendor and test files

cybertron:
These files either come from other projects or are not used in production so there's no reason to security scan them.


cybertron:
/skip

This is not a functional change, let's not waste ci resources testing it.

cybertron:
/override ci/prow/e2e-metal-ipi-ovn-ipv6

As noted above, this doesn't affect the e2e jobs, which failed on some image build error unrelated to anything in the repo.

PR #297, opened 2024-01-11
NO-ISSUE: Upgrade k8s api and client-go deps

mkowalski:


cybertron:
/retest-required
/lgtm
/hold

It looks like the ipv6 job was making progress and just ran out of time. It's unlikely the timeout was caused by this change. Once the e2e jobs have passed we can remove the hold.

cybertron:
/retest-required

Failing in different ways every time. :-(

PR #298, opened 2024-02-07
OCPBUGS-29095: Ignore egress IP when selecting Node IP

mkowalski:
This PR extends the logic of validating IP addresses whether suitable to be used as Node IP by adding a case when an IP address is marked as Egress IP (by e.g. OpenShift SDN). Those should never be used as Node IPs, thus we ignore them.


cybertron:
/lgtm

I feel like there may be shenanigans happening here, but it won't hurt to filter out egress ips anyway so let's do this.

PR #302, opened 2024-03-12
WIP: Skip collecting VRRP config on Cloud Platforms

sadasu:
GetVRRPConfig() provides information obtained from the node-ip service that currently run only on on-prem platforms. Added a boolean flag to getNodeConfig() so that VRRP config is not added to Node config on Cloud Platforms.

Another way to solve this problem would have been to check for the platform type within GetVRRPConfig() or getInterfaceAndNonVIPAddr(). That would require the vendored openshift/installer repo to be updated to have knowledge of all the currently supported platform types.

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -454,7 +454,7 @@ func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVips
 		} else {
 			ingressVip = nil
 		}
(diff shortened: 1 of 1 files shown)


sadasu:
The second option mentioned above has been implemented in https://github.com/openshift/baremetal-runtimecfg/pull/303

sadasu:
/close

The second option mentioned above has been implemented in #303

PR #310, opened 2024-05-07
Revert "OCPBUGS-32348: Make log level configurable"

mkowalski:
This reverts commit 0561ac232cd8c56913173fd14fb855376bcbbaf8.


mkowalski:
/cc @cybertron 

We should revert this change, I merged it before I realized it won't be as simple as passing env variable

PR #313, opened 2024-05-15
OCPBUGS-32141: Handle scenario when VIP does not belong to L2

mkowalski:
Lately we were made aware that the current logic of maching interfaces to VIPs is broken in vSphere IPI scenarios when VIP does not belong to Machine Network. Usually such a combination would be forbidden, but for vSphere it is allowed.

As a result, we may have e.g. an interface 10.8.38.29, machine network 10.8.42.0/23 and VIP 10.8.0.83. As counterintuitively it looks for baremetal, with vSphere it is allowed.

This PR fixes the logic we introduced in January 2023 where we added a hard-check to make sure VIP belongs to a subnet of the NIC. With this new logic, if the scenario described above happens, we will fallback to the logic from years ago when we were simply choosing a NodeInternalIP of the matching IP stack.

Closes: OCPBUGS-32141

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -605,22 +605,41 @@ func getSortedBackends(kubeconfigPath string, readFromLocalAPI bool, vips []net.
 	// by detecting which of the local interfaces belongs to the same subnet as requested VIP.
 	// This interface can be used to detect what was the original machine network as it contains
 	// the subnet mask that we need.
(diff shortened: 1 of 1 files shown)


cybertron:
/jira refresh
/label acknowledge-critical-fixes-only

This is a fix for a common use case on vsphere that is now being found in the field. We can't wait for 4.16 to ship before we get it in.

PR #315, opened 2024-05-27
OCPBUGS-32141: Handle scenario when VIP does not belong to L2 (addn. for ingress)

mkowalski:
In https://github.com/openshift/baremetal-runtimecfg/pull/313 we have fixed logic for topologies where VIP does not belong to the L2 of the node. But this change was missing additional part of the code that had the same flaw, i.e. generation of ingress config.

This PR is replicating the fix #313.

Fixes: OCPBUGS-32141

PR #316, opened 2024-05-29
OCPBUGS-32141: Fix handling of ELB Node IP detection

mkowalski:
In https://github.com/openshift/baremetal-runtimecfg/pull/313 we have fixed some logic regarding to VIPs in ELB scenario, but we forgot to handle the error. Because of this, even if we detected the IP correctly, the function could stil return a non-empty error causing the installation to fail.

This PR contains a simple fix to return `nil` instead of `err` in case we managed to recover from the issue.

A second commit fixes the issue where the change was missing additional part of the code that had the same flaw, i.e. generation of ingress config.

Fixes: OCPBUGS-32141

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -646,7 +646,7 @@ func getSortedBackends(kubeconfigPath string, readFromLocalAPI bool, vips []net.
 	sort.Slice(backends, func(i, j int) bool {
 		return backends[i].Address < backends[j].Address
 	})
-	return backends, err
+	return backends, nil
 }
 
 func GetLBConfig(kubeconfigPath string, apiPort, lbPort, statPort uint16, vips []net.IP) (ApiLBConfig, error) {
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -309,20 +309,44 @@ func GetIngressConfig(kubeconfigPath string, vips []string) (IngressConfig, erro
 		// by detecting which of the local interfaces belongs to the same subnet as requested VIP.
 		// This interface can be used to detect what was the original machine network as it contains
 		// the subnet mask that we need.
+		//
+		// In case there is no subnet containing a VIP on any of the available NICs we are counterintuitively
+		// selecting just a Node IP with the matching IP stack. This is a weird case in e.g. vSphere
+		// where VIPs do not belong to the L2 of the node, yet they work properly.
 		machineNetwork, err = utils.GetLocalCIDRByIP(vips[0])
-		if err != nil {
-			return ingressConfig, err
-		}
-	}
 
-	for _, node := range nodes.Items {
-		addr, err := getNodeIpForRequestedIpStack(node, vips, machineNetwork)
-		if err != nil {
+		if err == nil {
+			for _, node := range nodes.Items {
+				addr, err := getNodeIpForRequestedIpStack(node, vips, machineNetwork)
+				if err != nil {
+					log.WithFields(logrus.Fields{
+						"err": err,
+					}).Warnf("For node %s could not retrieve node's IP. Ignoring", node.ObjectMeta.Name)
+				} else {
+					ingressConfig.Peers = append(ingressConfig.Peers, addr)
+				}
+			}
+		} else {
 			log.WithFields(logrus.Fields{
 				"err": err,
-			}).Warnf("For node %s could not retrieve node's IP. Ignoring", node.ObjectMeta.Name)
-		} else {
-			ingressConfig.Peers = append(ingressConfig.Peers, addr)
+			}).Errorf("Could not retrieve subnet for IP %s. Falling back to an IP of the matching IP stack", vips[0])
+
+			for _, node := range nodes.Items {
+				addr := ""
+				for _, address := range node.Status.Addresses {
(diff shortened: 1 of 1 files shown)


cybertron (approved):
/lgtm

zhaozhanqi:
@mkowalski  Still met the keepalived is not ready on bootstrap and keepalived-monitor logs always show

```
time="2024-05-30T04:01:25Z" level=info msg="Failed to read ip from file /run/nodeip-configuration/ipv4" error="open /run/nodeip-configuration/ipv4: no such file or directory"
time="2024-05-30T04:01:25Z" level=warning msg="Could not retrieve ingress config: failed find a interface for the ip 192.168.222.40"
time="2024-05-30T04:01:26Z" level=info msg="Requested GetIngressConfig for empty VIP list."
time="2024-05-30T04:01:27Z" level=info msg="Requested GetIngressConfig for empty VIP list."
time="2024-05-30T04:01:28Z" level=info msg="Requested GetIngressConfig for empty VIP list."
```

mkowalski:
Yeah, I feel you need both #315 and this one here, #316... I will add a commit here and close #315

cybertron:
/retest-required

Seems to have failed on some sort of infrastructure issue: "Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist"

mkowalski:
Hey @zhaozhanqi, with tests passing I think it's ready for pre-merge test

mkowalski:
// needs backporting back to 4.14; manually because https://github.com/openshift/baremetal-runtimecfg/pull/313 should be also backported

zhaozhanqi:
> Hey @zhaozhanqi, with tests passing I think it's ready for pre-merge test

Hi, Trying to setup cluster with apivip which is out of machinenetworkCIDR in vsphere cluster.  but the master cannot bootup due to failed to access the ignition file from 22623

mkowalski:
The log pasted is only INFO and does not seem related to the failure. Even more, it's not even coming from any of our components
```
[controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.
```

mkowalski:
```
the master cannot bootup due to failed to access the ignition file from 22623
machineNetworkCIDR is - cidr: 192.168.221.0/24
```

I can't really help debugging this particular issue, to me this whole topology is very smelly and you need someone from VMware to help. To me it looks like this - your machine belongs to some /24 and you configure VIP outside of this /24. Now you get error that you cannot connect to the VIP.

From a pure networking perspective, your VIP is behind a router because it's not in the same /24. So unless VMware does some magic somewhere, it is not supposed to work.

I know this is topology people used in the past and it somehow magically works, but I never got my hands on a proper explanation why it works. So again, you need to ask people who claimed this is the way to go.

From my perspective - it does not work because nothing routes to VIP and our components never deploy routing components.

zhaozhanqi:
> I can't really help debugging this particular issue, to me this whole topology is very smelly and you need someone from VMware to help. To me it looks like this - your machine belongs to some /24 and you configure VIP outside of this /24. Now you get error that you cannot connect to the VIP.

Yes, this also confused to me.  Can we limited customer to use the VIP in same subnet with machineNetworkCIDR? 

> From a pure networking perspective, your VIP is behind a router because it's not in the same /24. So unless VMware does some magic somewhere, it is not supposed to work.

hmm, Actually, the two subnets 192.168.221.0/24 and 192.168.222.0/24 are routeable.  means if I have two vms in those two subnet,  and they can be ping each other.  but the VIP is not.  I don't have too knowledge on this and why ?

mkowalski:
> Can we limited customer to use the VIP in same subnet with machineNetworkCIDR?

No, @cybertron told me this is a relatively popular topology and we must keep having it

>Actually, the two subnets 192.168.221.0/24 and 192.168.222.0/24 are routeable. means if I have two vms in those two subnet, and they can be ping each other. but the VIP is not. I don't have too knowledge on this and why ?

As said, I don't know details and how this ever worked on vSphere but I am not sure if being routable is sufficient enough. Keepalived VIPs use ARP as an underlying protocol and this protocol is not routable (as opposed to ICMP). So there must be something at the vSphere layer that accepts the ARP announcements sent by node that holds VIP and then makes sure they are routed. And from your errors it looks to me like this "something at the vSphere layer" failed.

cybertron:
Yes, at some point we added validations to ensure the VIPs were part of the machine network and it broke a whole bunch of people on VSphere. It turns out that a common deployment method there is to use a default machine network and just use the actual network for the VIPs. I don't know how it works and I was surprised that it did.

I do know at one point we had a VSphere customer who had to open the firewall on their network to allow unicast VRRP traffic between subnets. It's possible something like that is needed here.

zhaozhanqi:
After this https://issues.redhat.com/browse/DPP-14984 we can setup the cluster on same vlan with 
```
  machineNetwork:
  - cidr: 10.184.15.128/25
  - cidr: 10.185.104.0/25
  networkType: OVNKubernetes
  serviceNetwork:
  - 172.30.0.0/16
platform:
  vsphere:
    apiVIPs:
    - 10.185.104.100
    ingressVIPs:
    - 10.185.104.101
```
and haproxy running well

PR #317, opened 2024-06-04
OCPBUGS-34706: Add support for OVN HostCidrs annotation

mkowalski:
This PR adds logic for extracting IP addresses for the Node from a newly added OVN annotation `k8s.ovn.org/host-cidrs`.

Till now we were only using the `k8s.ovn.org/host-addresses` annotation which is now deprecated in favour of the one mentioned above.

Fixes: OCPBUGS-34706

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -349,7 +349,8 @@ func getNodeIpForRequestedIpStack(node v1.Node, filterIps []string, machineNetwo
 	//
 	// We will use here the following sources:
 	//   1) Node.Status.Addresses list
-	//   2) Node annotation "k8s.ovn.org/host-addresses" in combination with Machine Networks
+	//   2) Node annotation "k8s.ovn.org/host-cidrs" in combination with Machine Networks
+	//   3) Deprecated node annotation "k8s.ovn.org/host-addresses" in combination with Machine Networks
 	//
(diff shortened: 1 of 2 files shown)


pkg/config/node.go
```diff
@@ -367,10 +368,24 @@ func getNodeIpForRequestedIpStack(node v1.Node, filterIps []string, machineNetwo
 		log.Debugf("For node %s can't find address using NodeInternalIP. Fallback to OVN annotation.", node.Name)
 
 		var ovnHostAddresses []string
-		if err := json.Unmarshal([]byte(node.Annotations["k8s.ovn.org/host-addresses"]), &ovnHostAddresses); err != nil {
+		var tmp []string
+
+		err := json.Unmarshal([]byte(node.Annotations["k8s.ovn.org/host-cidrs"]), &tmp)
```
rbbratta, line 373: I'm not sure how versioning works, is this going to work with upgrades?  4.13 to 4.14 maybe.  I don't know exactly when the annotation changed.

rbbratta: ah, I didn't have it on side-by-side diff.  🤣

rbbratta:
looks good on 4.16, testing upgrades.

cybertron (approved):
/lgtm

PR #327, opened 2024-08-13
OCPBUGS-38281: Handle errors in iptables healthcheck

cybertron:
Previously we ignored errors from the exists call in the firewall rule check. This can cause problems if the check fails for reasons other than the non-existence of the rule, such as a permission error on iptables. If the check returns err we should handle that and report the problem rather than just report that the firewall rule doesn't exist (which may or may not be true).


cybertron:
/retest-required

Deployment succeeded fine but the e2e tests timed out. I don't see anything in the monitor logs to suggest it was related to this patch.

PR #341, opened 2025-02-07
OCPBUGS-48469: Include cloud platform specific parameters to argument list for corednsmonitor 

sadasu:
Include cloud platform specific parameters to argument list for corednsmonitor

### DIFF
--- a/cmd/corednsmonitor/corednsmonitor.go
+++ b/cmd/corednsmonitor/corednsmonitor.go
@@ -77,6 +77,10 @@ func main() {
 	rootCmd.Flags().IPSlice("api-vips", nil, "Virtual IP Addresses to reach the OpenShift API")
 	rootCmd.Flags().IP("ingress-vip", nil, "DEPRECATED: Virtual IP Address to reach the OpenShift Ingress Routers")
 	rootCmd.Flags().IPSlice("ingress-vips", nil, "Virtual IP Addresses to reach the OpenShift Ingress Routers")
+	rootCmd.Flags().IPSlice("cloud-ext-lb-ips", nil, "IP Addresses of Cloud External Load Balancers for OpenShift API")
+	rootCmd.Flags().IPSlice("cloud-int-lb-ips", nil, "IP Addresses of Cloud Internal Load Balancers for OpenShift Internal API")
+	rootCmd.Flags().IPSlice("cloud-ingress-lb-ips", nil, "IP Addresses of Cloud Ingress Load Balancers")
+
 	if err := rootCmd.Execute(); err != nil {
 		log.Fatalf("Failed due to %s", err)
 	}


mkowalski:
/test e2e-metal-ipi-ovn-ipv6

This one has to pass

pkg/config/node.go
```diff
@@ -487,7 +487,17 @@ func getNodeIpForRequestedIpStack(node v1.Node, filterIps []string, machineNetwo
...
+	if clusterConfigPath != "" {
+		if onPremPlatform, err = isOnPremPlatform(clusterConfigPath); err != nil {
+			return Node{}, err
+		}
+	} else {
+		if len(clusterLBConfig.ApiIntLBIPs) > 0 {
+			onPremPlatform = false
+		}
+	}
+	if !onPremPlatform {
```
mkowalski, line 500: I have to admit, I have a hard time understanding this logic. Correct me if I read it wrong:

- If clusterConfigPath is accessible, you determine the platform by calling isOnPremPlatform function
- If clusterConfigPath is NOT accessible, if at least 1 ApiIntLBIPs is provided, the platform is NOT on-prem

If this is what you wanted, why is that? What do I miss so that existence of ApiIntLBIPs is enough to say that you are NOT on-prem? What will happen if I take `platform: baremetal` with ApiIntLBIPs but without install-config?

mkowalski (changes requested):

pkg/config/node.go
sadasu: `BareMetal` and other on-prem platforms only sets the `api-vips` field. The `ApiIntLBIPs` are set by https://github.com/openshift/baremetal-runtimecfg/blob/main/cmd/runtimecfg/render.go#L37 only by cloud platforms so I felt confident to use that field to determine the platform type. 
I agree that approximation is not ideal.

So, now I have updated PR to bring in another parameter,  "--platform-type" to render. This parameter is currently passed in only when running on cloud platforms.

pkg/config/node.go
```diff
@@ -487,7 +489,17 @@ func getNodeIpForRequestedIpStack(node v1.Node, filterIps []string, machineNetwo
...
+	onPremPlatform := true
+	if clusterConfigPath != "" {
+		if onPremPlatform, err = isOnPremPlatform(clusterConfigPath); err != nil {
+			return Node{}, err
+		}
+	} else {
+		if clusterLBConfig.PlatformType == string(configv1.GCPPlatformType) || clusterLBConfig.PlatformType == string(configv1.AWSPlatformType) {
+			onPremPlatform = false
+		}
+	}
```
mkowalski, line 501: Let's try NOT modifying this piece of code because it makes the whole function `isOnPremPlatform()` severely defunct. Instead, let's try the following

1) change signature of `isOnPremPlatform` to `isOnPremPlatform(configPath string, platformType string) (bool, error)`
2) modify the body of `isOnPremPlatform` to something like

```
        if platformType == "baremetal" || platformType == "vsphere" || [...] {
		return true, nil
	} <---- THIS IS NEW

	ic, err := getClusterConfigMapInstallConfig(configPath)
	if err != nil {
		return true, err
	}
	if ic.Platform.BareMetal != nil || ic.Platform.VSphere != nil || ic.Platform.OpenStack != nil || ic.Platform.Ovirt != nil || ic.Platform.Nutanix != nil {
		return true, nil
	}
	return false, nil
}
```
3) wherever we call `isOnPremPlatform`, add your newly created `--platform-type` parameter

I am pushing hard for this, because without this (i.e. in how this is written right now) the function `isOnPremPlatform` will return WRONG value if for any reason you run runtimecfg with install-config for Metal but cmdline param with a value of e.g. "AWS". It may happen for any random reason (e.g. bug in MCO) but we simply cannot trust that every passed set of cmdline params is correct (because people hack stuff in various ways).

Basically the whole reason for asking to introduce `--platform-type` was to make `isOnPremPlatform` function smarter.

pkg/config/node.go
```diff
@@ -91,6 +92,7 @@ type ClusterLBConfig struct {
 	ApiLBIPs     []net.IP
 	ApiIntLBIPs  []net.IP
 	IngressLBIPs []net.IP
+	PlatformType string
```
mkowalski, line 95: I would rather see `PlatformType string` defined inside `type Cluster struct` because platform is a property of the cluster, not of the loadbalancer

pkg/config/node.go
```diff
@@ -521,7 +533,7 @@ func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVips
 	return nodes[0], nil
 }
 
-func getNodeConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVip net.IP, ingressVip net.IP, apiPort, lbPort, statPort uint16) (node Node, err error) {
+func getNodeConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVip net.IP, ingressVip net.IP, apiPort, lbPort, statPort uint16, onPremPlatform bool) (node Node, err error) {
```
mkowalski, line 536: If possible, let's not rely on the computed state and passed value of `onPremPlatform bool`, because by not handling the error from `if onPremPlatform, err := isOnPremPlatform(clusterConfigPath); !onPremPlatform {` we will never notice if someone called `getNodeConfig` without thinking if they correctly pass `true` or `false`.

Calling `isOnPremPlatform` inside `getNodeConfig` is not expensive so I'd rather keep it there. If in the future, for any reason we realize there is a performance bottleneck inside this function, then we will handle it

sadasu: Done.

pkg/config/node.go
sadasu: Agreed. I think I rushed through this initially.

patrickdillon:
I think I don't understand the motivation/context. If there is existing code for determining whether the platform is onPrem or not, as there seems to be, but that code doesn't work in the cloud custom dns use case because the path to the cluster-config is not passed, why not fix this by passing the path?

That would seem simpler than adding an additional parameter and code to determine the platform. Can we not do that?

I could not find any relevant details in the attached bug. Also [all the references to `--cloud-config` I could find in MCO ](https://github.com/search?q=repo%3Aopenshift%2Fmachine-config-operator%20--cluster-config&type=code) do not seem conditional, so I feel like I'm missing something.

sadasu:
> Also [all the references to `--cloud-config` I could find in MCO ](https://github.com/search?q=repo%3Aopenshift%2Fmachine-config-operator%20--cluster-config&type=code) do not seem conditional, so I feel like I'm missing something.

Those references are for the CoreDNS pod that is running on the bootstrap node.

sadasu:
> I think I don't understand the motivation/context. If there is existing code for determining whether the platform is onPrem or not, as there seems to be, but that code doesn't work in the cloud custom dns use case because the path to the cluster-config is not passed, why not fix this by passing the path?
> 
I am not sure why the CoreDNS pod for on-prem platforms does not pass in the path to cluster-config.

For the cloud-platform case, I had 2 choices:
1. Volume mount the path to cluster config so that it is available within the container (assuming it is available on the host). Pass path to cluster-config as a parameter to the respective containers.
2. Pass the platformType as a new parameter because that is the only value we are interested in within the cluster-config. 

I went with Option 2 because the value of Platform type does not change during the life of the cluster whereas the other files that are volume mounted are maintained by other system.d units which means that they could technically change during the life of the container.

I see going with Option 1 would maintain consistency with what is happening on the bootstrap node.

@mkowalski, WDYT?

sadasu:
Oops! closed the PR by mistake.

mkowalski:
>That would seem simpler than adding an additional parameter and code to determine the platform. Can we not do that?

>For the cloud-platform case, I had 2 choices:
>Volume mount the path to cluster config so that it is available within the container (assuming it is available on the host). Pass path to cluster-config as a parameter to the respective containers.
>Pass the platformType as a new parameter because that is the only value we are interested in within the cluster-config.

My take here is that option 1 is desired, but for some reason it does not work as you want. Basically, according to the MCO the path should be passed, but when you run the system, you are not getting the path.

This should be debugged. We agreed on accepting option 2 as a quick&dirty way to solve it because we were always told that it is critical to have this feature working. But this is not a desired way.

The desired way is to debug the issue, find out why the file is not passed as we expect it, and fix that.

sadasu:
> > That would seem simpler than adding an additional parameter and code to determine the platform. Can we not do that?
> 
> > For the cloud-platform case, I had 2 choices:
> > Volume mount the path to cluster config so that it is available within the container (assuming it is available on the host). Pass path to cluster-config as a parameter to the respective containers.
> > Pass the platformType as a new parameter because that is the only value we are interested in within the cluster-config.
> 
> My take here is that option 1 is desired, but for some reason it does not work as you want. Basically, according to the MCO the path should be passed, but when you run the system, you are not getting the path.
> 
> This should be debugged. We agreed on accepting option 2 as a quick&dirty way to solve it because we were always told that it is critical to have this feature working. But this is not a desired way.
> 
> The desired way is to debug the issue, find out why the file is not passed as we expect it, and fix that.

Ok. I missed the fact that platform-type was considered the quick and dirty way. I somehow thought that was preferred. Of course, we can pass in the install-config by mounting the appropriate volume. Let me rework to that.

PR #343, opened 2025-02-26
OPNET-629: Use HAProxy monitor endpoint instead of API

cybertron:
Same as #336 which was reverted because it merged before a dependency in MCO.


cybertron:
/retest
/hold cancel

https://github.com/openshift/machine-config-operator/pull/4767 has merged, we should be good to get this in.

cybertron:
/retest

Clusters deployed but the e2e tests timed out. Looking at logs it appears this patch is working so I don't think it's related.

mkowalski:
/lgtm

/hold
Only because it's old and I don't know if you want to merge it now. Remove as you wish

PR #344, opened 2025-03-16
OCPEDGE-1689: Include Arbiter nodes in the API backends if there is only one master

giladravid16:
Following the [enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/arbiter-clusters.md) for TNA clusters, we want to install this new type of cluster using Assisted-Service.

Currently there is a problem when installing TNA clusters using Assisted-Service, because in Assisted-Service one of the master nodes acts as the bootstrap. So during the installation there will only be one master node, but we need two in order for keepalived-monitor to configure keepalived.
We cannot wait until the bootstrap finishes and becomes a master, because then no node will have the API vip.
To circumvent that we will add the arbiter node to the list of nodes.

### DIFF
--- a/pkg/config/node.go
+++ b/pkg/config/node.go
@@ -643,6 +643,22 @@ func getSortedBackends(kubeconfigPath string, readFromLocalAPI bool, vips []net.
 		}).Info("Failed to get master Nodes list")
 		return []Backend{}, err
 	}
+	// When installing TNA clusters using assisted service, one of the master nodes acts as the bootstrap.
+	// So during the installation there will only be one master node, but we need two in order to configure keepalived.
+	// We cannot wait until the bootstrap finishes and becomes a master, because then no node will have the API vip.
+	// To circumvent that we will add the arbiter node to the list of nodes.
+	if len(nodes.Items) == 1 {
+		arbiters, err := clientset.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
+			LabelSelector: "node-role.kubernetes.io/arbiter=",
+		})
+		if err != nil {
+			log.WithFields(logrus.Fields{
+				"err": err,
+			}).Info("Failed to get arbiter Nodes list")
+			return []Backend{}, err
+		}
+		nodes.Items = append(nodes.Items, arbiters.Items...)
+	}
 	if len(vips) == 0 {
 		return []Backend{}, fmt.Errorf("Trying to build config using empty VIPs")
 	}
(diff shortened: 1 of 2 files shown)


mkowalski:
Looking at the PR, this looks like enabling keepalived for 2 node cluster that is a 1 node cluster in the transient state. Is there enhancement for this? Can you please link? Will proceed with review once we agree that this is topology that OCP supports

/hold

pkg/monitor/dynkeepalived.go
```diff
@@ -100,6 +100,7 @@ func doesConfigChanged(curConfig, appliedConfig *config.Node) bool {
 	// approach we should avoid asymetric configuration
 	if curConfig.EnableUnicast {
 		if os.Getenv("IS_BOOTSTRAP") == "no" && len(curConfig.LBConfig.Backends) < 2 {
+			log.Warnf("Invalid keepalived.conf - number of api backends must be at least 2 got %d", len(curConfig.LBConfig.Backends))
```
mkowalski, line 103: nit: this function technically is not specific to API, it can be as well used for any arbitrary type of backend; let's keep the error message generic, scratch the word "api"

pkg/config/node.go
```diff
@@ -643,6 +643,22 @@ func getSortedBackends(kubeconfigPath string, readFromLocalAPI bool, vips []net.
 		}).Info("Failed to get master Nodes list")
 		return []Backend{}, err
 	}
+	// When installing TNA clusters using assisted service, one of the master nodes acts as the bootstrap.
+	// So during the installation there will only be one master node, but we need two in order to configure keepalived.
+	// We cannot wait until the bootstrap finishes and becomes a master, because then no node will have the API vip.
+	// To circumvent that we will add the arbiter node to the list of nodes.
+	if len(nodes.Items) == 1 {
```
cybertron, line 650: Is keepalived always running on the arbiter? If so, it should always be included in the peer list. Otherwise you run the risk of split brain where the arbiter has keepalived sending announcement packets but the other nodes don't know about it and can't coordinate properly.

giladravid16:
@mkowalski I linked the openshift enhancement for TNA clusters in the description.
Basically those clusters can have 2 masters plus an extra node just for components like etcd that require quorum (but don't run the kube-api for example). The problem this PR solves is specifically for installing TNA clusters using Assisted-Service because we use 1 of the master nodes as the bootstrap, so during the installation the api-vip needs to move from the bootstrap to the other master node while that node is the only master in the cluster. At the end of the installation the bootstrap node will become a master and the keepalived.conf will include only the 2 masters in the api-vip backend.

PR #345, opened 2025-03-19
OCPBUGS-48469: Pass in platform as a optional parameter to runtimecfg cmds

sadasu:
When running on master nodes, cluster-config.yaml is not available to lookup the platform and determine if it is an on-prem or cloud platform.
Fixing this by adding  platform as an optional parameter. When platform and cluster-config.yaml are both not present, we will assume on-prem platform to preserve backwards compatibility.


cmd/corednsmonitor/corednsmonitor.go
```diff
@@ -67,8 +67,9 @@ func main() {
 			if err != nil {
 				cloudIngressLBIPs = []net.IP{}
 			}
+			platformType, err := cmd.Flags().GetString("platform-type")
```
mkowalski, line 70: Should there be `if err != nil {` after this line?

pkg/config/node.go
```diff
@@ -160,7 +162,10 @@ func getClusterConfigMasterAmount(configPath string) (amount *int64, err error)
 	return ic.ControlPlane.Replicas, nil
 }
 
-func isOnPremPlatform(configPath string) (bool, error) {
+func isOnPremPlatform(configPath string, platformType string) (bool, error) {
+	if platformType == string(configv1.GCPPlatformType) || platformType == string(configv1.AWSPlatformType) {
+		return false, nil
+	}
```
mkowalski, line 168: 1) What happens if you pass `ic.Platform.BareMetal` and `platformType configv1.GCPPlatformType`? There should be a check throwing a hard error if install-config and platformType do not match

2) What happens if you pass `ic.Platform.Aws` and `platformType configv1.BareMetalPlatformType`?

This function now looks like it handles half of stuff one way and half of stuff the other. Just for completeness and for having the clear code, it should do it the same way always.

In https://github.com/openshift/machine-config-operator/pull/4830 you are adding `platformType` for whoever consumes `templates/common/cloud-platform-alt-dns` (in runtimecfg we don't care much about who that is). That means we have the following possibilities of runtimecfg

1) platformType exists, install-config does not
2) platformType exists, install-config exists
3) platformType does not exist, install-config exists
4) platformType does not exist, install-config does not exist

Out of those four options, (1) and (3) are easy because there can be no conflict. Option (4) is actually a problem and we don't care about handling it in this function here. For option (2), where both value exist, we need to be explicit that both match, otherwise it's a question of time that someone messes up

cmd/corednsmonitor/corednsmonitor.go
```diff
@@ -80,6 +81,7 @@ func main() {
 	rootCmd.Flags().IPSlice("cloud-ext-lb-ips", nil, "IP Addresses of Cloud External Load Balancers for OpenShift API")
 	rootCmd.Flags().IPSlice("cloud-int-lb-ips", nil, "IP Addresses of Cloud Internal Load Balancers for OpenShift Internal API")
 	rootCmd.Flags().IPSlice("cloud-ingress-lb-ips", nil, "IP Addresses of Cloud Ingress Load Balancers")
+	rootCmd.Flags().String("platform-type", "", "PlatformType of cloud")
```
mkowalski, line 84: doc nit: this field is not about a cloud, it's just a platform type so more appropriate description is something like `platformType provides the platform name`

cmd/runtimecfg/display.go
```diff
@@ -31,6 +31,7 @@ func init() {
 	displayCmd.Flags().IPSlice("cloud-ext-lb-ips", nil, "IP Addresses of Cloud External Load Balancers for OpenShift API")
 	displayCmd.Flags().IPSlice("cloud-int-lb-ips", nil, "IP Addresses of Cloud Internal Load Balancers for OpenShift API")
 	displayCmd.Flags().IPSlice("cloud-ingress-lb-ips", nil, "IP Addresses of Cloud Ingress Load Balancers")
+	displayCmd.Flags().String("platform-type", "", "Optional PlatformType of cloud")
```
mkowalski, line 34: doc nit: this field is not about a cloud, it's just a platform type so more appropriate description is something like `platformType provides the platform name`

cmd/runtimecfg/render.go
```diff
@@ -36,6 +36,7 @@ func init() {
 	renderCmd.Flags().IPSlice("cloud-ext-lb-ips", nil, "IP Addresses of Cloud External Load Balancers for OpenShift API")
 	renderCmd.Flags().IPSlice("cloud-int-lb-ips", nil, "IP Addresses of Cloud Internal Load Balancers for OpenShift Internal API")
 	renderCmd.Flags().IPSlice("cloud-ingress-lb-ips", nil, "IP Addresses of Cloud Ingress Load Balancers")
+	renderCmd.Flags().String("platform-type", "", "Optional PlatformType of cloud")
```
mkowalski, line 39: doc nit: this field is not about a cloud, it's just a platform type so more appropriate description is something like `platformType provides the platform name`

pkg/config/node.go
```diff
@@ -486,10 +491,10 @@ func getNodeIpForRequestedIpStack(node v1.Node, filterIps []string, machineNetwo
 // lbPort: The port on which haproxy listens.
 // statPort: The port on which the haproxy stats endpoint listens.
 // clusterLBConfig: A struct containing IPs for API, API-Int and Ingress LBs
-func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVips, ingressVips []net.IP, apiPort, lbPort, statPort uint16, clusterLBConfig ClusterLBConfig) (node Node, err error) {
-	if onPremPlatform, _ := isOnPremPlatform(clusterConfigPath); !onPremPlatform {
+func GetConfig(kubeconfigPath, clusterConfigPath, resolvConfPath string, apiVips, ingressVips []net.IP, apiPort, lbPort, statPort uint16, clusterLBConfig ClusterLBConfig, platformType string) (node Node, err error) {
```
mkowalski, line 494: missing docstring about `platformType` in a line above

pkg/monitor/dynkeepalived.go
```diff
@@ -355,7 +355,7 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
 
 		case desiredModeInfo := <-updateModeCh:
 
-			newConfig, err := config.GetConfig(kubeconfigPath, clusterConfigPath, "/etc/resolv.conf", apiVips, ingressVips, 0, 0, 0, config.ClusterLBConfig{})
+			newConfig, err := config.GetConfig(kubeconfigPath, clusterConfigPath, "/etc/resolv.conf", apiVips, ingressVips, 0, 0, 0, config.ClusterLBConfig{}, "")
```
mkowalski, line 358: Unless there is a reason not to, we should keep passing the `platform` parameter all around instead of placeholder with empty value

pkg/monitor/dynkeepalived.go
```diff
@@ -430,7 +430,7 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
 				// if the path doesn't exist then RemoveAll returns nil
 				log.WithFields(logrus.Fields{"path": iptablesFilePath}).WithError(err).Error("Failed to remove file")
 			}
-			newConfig, err := config.GetConfig(kubeconfigPath, clusterConfigPath, "/etc/resolv.conf", apiVips, ingressVips, 0, 0, 0, config.ClusterLBConfig{})
+			newConfig, err := config.GetConfig(kubeconfigPath, clusterConfigPath, "/etc/resolv.conf", apiVips, ingressVips, 0, 0, 0, config.ClusterLBConfig{}, "")
```
mkowalski, line 433: Unless there is a reason not to, we should keep passing the `platform` parameter all around instead of placeholder with empty value

pkg/config/node.go
sadasu: Option 4 actually happens on on-prem master nodes. Ignition on master nodes does not lay down install-config. So, without platform parameter and install-config, original logic defaults this case to on-prem platforms (without this the e2e-metal-ipi-ovn-ipv* jobs will not pass).

Also, do we need to check if platforms provided by "platform" parameter and install config actually match or is it sufficient of they both point to on-prem or cloud?

pkg/monitor/dnsmasqmonitor.go
```diff
@@ -33,7 +33,7 @@ func DnsmasqWatch(kubeconfigPath, templatePath, cfgPath string, apiVips []net.IP
 			return nil
 		default:
 			// We only care about the api vip and cluster domain here
-			config, err := config.GetConfig(kubeconfigPath, "", "/etc/resolv.conf", apiVips, apiVips, 0, 0, 0, config.ClusterLBConfig{})
+			config, err := config.GetConfig(kubeconfigPath, "", "/etc/resolv.conf", apiVips, apiVips, 0, 0, 0, config.ClusterLBConfig{}, "")
```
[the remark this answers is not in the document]
sadasu: DnsmasqWatch() is called on on on-prem platforms. Do you want me to add "platform" as an optional  parameter to the `dnsmasqmonitor` command?

pkg/monitor/dynkeepalived.go
sadasu: Same case as https://github.com/openshift/baremetal-runtimecfg/pull/345/files#r2035844045. I am trying not to introduce too many changes within code that is run only on on-prem environments.

patrickdillon:
Do we really need to know which platform is running here? I'm trying to understand what difference platform makes, and it seems that it determines whether the clusterLBConfig is used. Can we just switch that behavior based on whether clusterLBConfig is populated and forget about platform? That is, if clusterLBConfig is populated, use it.

pkg/monitor/dnsmasqmonitor.go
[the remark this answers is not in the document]
mkowalski: Yes please. In the past baremetal-runtimecfg was called only on on-prem platforms yet we are where we are and every platform now uses it. It will be the same, sooner or later, with this function call here.

pkg/config/node_test.go
```diff
@@ -338,6 +343,145 @@ func deleteTempResolvConf() {
...
+			Name:      "cluster-config-v1",
+			Namespace: "kube-system",
+		},
+		Data: map[string]string{
+			"install-config": string(icData),
+		},
+	}
+
+	controllerConfig, _ := yaml.Marshal(cm)
+	os.WriteFile("/tmp/install-config.yaml", controllerConfig, 0644)
```
mkowalski, line 403: Let's use `os.MkdirTemp()` instead of hardcoding `/tmp`. It won't necessary always be `/tmp` so tests will not work if ran in such an environment

mkowalski:
>Do we really need to know which platform is running here? I'm trying to understand what difference platform makes [...]

Yeah this is something I was vouching for. Basically, we reached the stage where there is too much magic and guesstimating happening around, stuff like "if we can't detect the platform but we have at least 1 VIP provided and it was not raining for the past 4 days, then it must be bare metal; if the same but it was raining yesterday, then it's vSphere"

We started as runtimecfg being used purely for bare metal, so there was no need for passing platform, because it was only used for one. Then we started using it wider, for a set of on-prem platforms, not necessarily bare metal but platforms that require slightly different behaviour (like for example, OpenStack is very much on-prem platform but not really bare metal). Then we went completely sideways because we started using something called "baremetal-runtimecfg" on AWS/GCP which are neither metal nor on-prem platforms. Most of this time, we did not care to explicitly say which platform is consuming the project.

I agree maybe we could again guesstimate that if platform is not explicitly passed and some conditions are satisfied by clusterLBConfig and some other conditions are satisfied by clusterYetAnotherConfig then we are on GCP, but that starts to be too complex to understand.

I feel at some point we have to stop it and force everyone to always call runtimecfg with explicit platform. I feel right now is this point

pkg/config/node_test.go
sadasu: Done.

cmd/corednsmonitor/corednsmonitor.go
sadasu: Yes. Fixed.

pkg/monitor/dynkeepalived.go
sadasu: Done.

pkg/monitor/dynkeepalived.go
sadasu: Done.

pkg/monitor/dnsmasqmonitor.go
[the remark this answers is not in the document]
sadasu: Done.

cmd/runtimecfg/render.go
sadasu: Done.

pkg/config/node.go
sadasu: Done.

cmd/runtimecfg/display.go
sadasu: Fixed.

pkg/config/node_test.go
```diff
@@ -338,6 +343,148 @@ func deleteTempResolvConf() {
 	os.Remove("/tmp/resolvConf")
 }
 
+var (
+	icFilePath    = "/tmp/install-config.yaml"
```
mkowalski, line 347: Still hardcoded `/tmp` here

sadasu: That was just the initial value that was updated after the file was created. Anyways, updated.

PR #389, opened 2026-04-28
NO-ISSUE: Bump go to fix security vulnerabilities

cybertron:
We should be using go 1.25 anyway in this release.

## Summary by CodeRabbit

* **Chores**
  * Removed explicit Go toolchain directive and updated the project's declared Go version to 1.25.8 to ensure consistent builds and toolchain compatibility.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/openshift/baremetal-runtimecfg
 
-go 1.24.0
-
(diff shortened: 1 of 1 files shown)


cybertron:
Unlinking from the bug because this is only a partial fix. We'll need a followup when the golang builder has 1.25.9 available.

/override ci/prow/security

PR #391, opened 2026-05-29
OCPBUGS-86571: node-ip: wait for both address families on dual-stack clusters

mkowalski:
## Summary

On RHEL 10, IPv4 ACD (Address Conflict Detection) is enabled by default with a ~200ms timeout. Unlike IPv6 DAD, IPv4 ACD **does not put addresses in tentative state** — it withholds them entirely from the interface until ARP probing completes.

This creates a race condition on dual-stack baremetal clusters:

```
T+0s  Interface up, IPv6 tentative, IPv4 ACD probing starts
T+2s  IPv6 DAD complete → address permanent
      nodeip-configuration.service starts
T+3s  Queries addresses → sees only IPv6 → writes IPv6-only config ❌
T+4s  IPv4 ACD complete → IPv4 added (too late)
```

Kubelet ends up with `--node-ip=<IPv6 only>` despite the interface eventually having both addresses.

## Root Cause

`getSuitableIPs()` declares success as soon as it finds **any** IP address. In a dual-stack cluster, finding just IPv6 is treated as success, so the retry loop never kicks in for the missing IPv4.

## Fix

Make `getSuitableIPs()` dual-stack-aware with a bounded wait:

1. **Auto-detect dual-stack from VIPs**: if the VIP list contains both IPv4 and IPv6 addresses, the cluster is dual-stack. This covers on-prem/baremetal deployments (the reported scenario).

2. **`--dual-stack` flag**: for non-VIP deployments (None/External platforms), the MCO can explicitly signal dual-stack. This is a new flag that the MCO template can pass via `{{if or (eq .IPFamilies "DualStack") (eq .IPFamilies "DualStackIPv6Primary")}}--dual-stack{{end}}`.

3. **Bounded 10s wait**: when dual-stack is detected but only one family is found, retry for up to 10 seconds. This covers the typical ACD delay (~200ms-3s) with margin, without blocking boot.

4. **Graceful fallback**: if the second family never appears, proceed with what was found and log a warning. No change in behavior for single-stack clusters.

## Why not disable IPv4 DAD entirely?

An alternative fix (PR [openshift/machine-config-operator#6098](https://github.com/openshift/machine-config-operator/pull/6098)) disables IPv4 DAD globally. However:
- CEE reports ~113 support cases in 3 years from duplicate IPv4 addresses
- IPv4 DAD was enabled in RHEL 10 specifically to address this
- Disabling it trades one class of problems for another

This fix preserves IPv4 DAD while making `nodeip-configuration` tolerant of the ACD delay.

## Testing

- All existing tests pass (50 address tests + config/nodeconfig/loggingconfig suites)
- Added unit tests for `vipsAreDualStack()`, `hasBothIPFamilies()`, and `shouldKeepWaitingForDualStack()`

Fixes: https://issues.redhat.com/browse/OCPBUGS-86571

---
🤖 *This PR was created by [OpenClaw](https://openclaw.ai) on behalf of @mkowalski.*

## Summary by CodeRabbit

* **New Features**
  * Added the `--dual-stack` option for selecting both IPv4 and IPv6 addresses.
  * Dual-stack mode now applies to VIP and default-route address selection.
  * Retry mode waits until both address families are available, with progress logging and periodic updates.
  * Without retry, the system warns when only one address family is available and may continue.
  * Improved IPv6 address validation and VIP selection consistency.

* **Tests**
  * Expanded coverage for dual-stack detection, retries, logging, backoff, and address discovery.

### DIFF
--- a/cmd/runtimecfg/node-ip.go
+++ b/cmd/runtimecfg/node-ip.go
@@ -28,6 +28,13 @@ const (
 	ovn                         = "OVNKubernetes"
 	maxSecondsToSuitableIPsLoop = 300 // 5 minutes
 	addSecondsToSuitableIPsLoop = 2
+
+	// maxDualStackWaitSeconds is the maximum additional time we wait for the
+	// second address family to appear on a dual-stack cluster. RHEL 10 enables
+	// IPv4 ACD (Address Conflict Detection) by default which delays IPv4
+	// address assignment by ~200ms-3s. During this window nodeip-configuration
+	// can see only IPv6 and incorrectly write a single-stack config.
+	maxDualStackWaitSeconds = 10
 )
 
 var nodeIPCmd = &cobra.Command{
@@ -69,6 +76,7 @@ var params struct {
 	userManagedLB bool
 	networkType   string
 	platform      string
+	dualStack     bool
 }
 
 // init executes upon import
@@ -80,6 +88,7 @@ func init() {
 	nodeIPCmd.PersistentFlags().StringVarP(&params.networkType, "network-type", "n", ovn, "CNI network type")
 	nodeIPCmd.PersistentFlags().BoolVarP(&params.userManagedLB, "user-managed-lb", "l", false, "User managed load balancer")
 	nodeIPCmd.PersistentFlags().StringVarP(&params.platform, "platform", "p", "", "Cluster platform")
+	nodeIPCmd.PersistentFlags().BoolVar(&params.dualStack, "dual-stack", false, "Cluster is configured as dual-stack (wait for both address families)")
 	rootCmd.AddCommand(nodeIPCmd)
 }
 
@@ -89,7 +98,8 @@ func show(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	chosenAddresses, _, err := getSuitableIPs(params.retry, vips, params.preferIPv6, params.networkType)
+	dualStack := params.dualStack || vipsAreDualStack(vips)
+	chosenAddresses, _, err := getSuitableIPs(params.retry, vips, params.preferIPv6, params.networkType, dualStack)
 	if err != nil {
 		return err
 	}
@@ -114,7 +124,8 @@ func set(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	chosenAddresses, matchesVips, err := getSuitableIPs(params.retry, vips, params.preferIPv6, params.networkType)
+	dualStack := params.dualStack || vipsAreDualStack(vips)
+	chosenAddresses, matchesVips, err := getSuitableIPs(params.retry, vips, params.preferIPv6, params.networkType, dualStack)
 	if err != nil {
 		return err
 	}
@@ -218,7 +229,7 @@ func checkAddressUsable(chosen []net.IP) (err error) {
 	return err
 }
 
-func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType string) (chosen []net.IP, matchesVips bool, err error) {
+func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType string, dualStack bool) (chosen []net.IP, matchesVips bool, err error) {
 	// timerLoop will hold a time in Seconds to be used with time.Sleep() before going
 	// for the next loop interation.
 	timerLoop := 1
@@ -227,6 +238,11 @@ func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType stri
 	if os.Getenv("ENABLE_NODEIP_DEBUG") == "true" {
 		utils.SetDebugLogLevel()
 	}
+
+	// dualStackWaitStart tracks when we first found a single-family result in
+	// a dual-stack cluster. We allow a bounded extra wait for the second family
+	// to appear (e.g. IPv4 delayed by ACD/DAD probing on RHEL 10).
+	var dualStackWaitStart time.Time
 
 	ipFilterFunc := utils.ValidNodeAddress
 	for {
@@ -243,6 +259,13 @@ func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType stri
 					}
 					time.Sleep(time.Second)
 					continue
+				}
+				if dualStack && !hasBothIPFamilies(chosen) && retry {
+					if shouldKeepWaitingForDualStack(&dualStackWaitStart, chosen) {
+						continue
+					}
+					log.Warnf("Dual-stack cluster but only found %d address(es) after %ds wait, proceeding with what we have: %v",
+						len(chosen), maxDualStackWaitSeconds, chosen)
 				}
 				return chosen, true, err
 			}
@@ -264,6 +287,14 @@ func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType stri
 					chosen = []net.IP{}
 					time.Sleep(time.Second)
 					continue
+				}
+				if dualStack && !hasBothIPFamilies(chosen) && retry {
+					if shouldKeepWaitingForDualStack(&dualStackWaitStart, chosen) {
+						chosen = []net.IP{}
+						continue
+					}
+					log.Warnf("Dual-stack cluster but only found %d address(es) after %ds wait, proceeding with what we have: %v",
+						len(chosen), maxDualStackWaitSeconds, chosen)
 				}
 				return chosen, false, err
 			}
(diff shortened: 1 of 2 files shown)


mkowalski:
/override ci/prow/security

Fixed by https://github.com/openshift/baremetal-runtimecfg/pull/392

cybertron:
@rbbratta Can you verify this?

cmd/runtimecfg/node-ip.go
```diff
@@ -294,6 +325,54 @@ func parseIPs(args []string) ([]net.IP, error) {
...
+			hasIPv4 = true
+		}
+	}
+	return hasIPv4 && hasIPv6
+}
+
+// shouldKeepWaitingForDualStack decides whether to retry when we found IPs from
+// only one address family in a dual-stack cluster. It implements a bounded wait
+// of maxDualStackWaitSeconds to handle IPv4 ACD delays (RHEL 10+) without
+// blocking boot indefinitely.
```
rbbratta, line 362: I'm confused.  I thought we wanted to wait forever in dual-stack case or how does this interact with https://github.com/openshift/machine-config-operator/pull/6289  

We need to write `/run/nodeip-configuration/ipv6` in the dual-stack case so what we wait after `nmstate`.

rbbratta: or https://github.com/openshift/machine-config-operator/pull/5350

cmd/runtimecfg/node-ip.go
```diff
@@ -89,7 +98,8 @@ func show(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	chosenAddresses, _, err := getSuitableIPs(params.retry, vips, params.preferIPv6, params.networkType)
+	dualStack := params.dualStack || vipsAreDualStack(vips)
+	chosenAddresses, _, err := getSuitableIPs(params.retry, vips, params.preferIPv6, params.networkType, dualStack)
```
rbbratta, line 102: This is gonna be a bit confusing because technically this could be `NODEIP_HINT`hints as well, right? And because we do not quote the `NODEIP_HINT` variable we could white space inject multiple node IP hints. 

I guess you could have IPv4 VIPs and IPv6 `NODEIP_HINT`  that would also trigger dual-stack.

rbbratta: This is also confusing because, well we need to work out the dual stack to single stack IPv6 conversion scenario, because conversion might get hung, because as far as I know the VIPs do not count when determining if a cluster is dual stack. I think we've specified if a cluster is single stack, it should not have the other addresses present, so single IPv4 should not have IPv6 VIP, or single IPv6 should not have IPv4 VIP?

rbbratta: Claude guesses:


 During dual-to-single conversion:

 1. Admin patches networks.config.openshift.io to remove the IPv6 CIDR
 2. MCO re-renders: IPFamilies = IPv4, --dual-stack flag is removed from nodeip-configuration.service
 3. But VIPs in infrastructure.config.openshift.io still contain IPv6 (separate API object, separate patch step)
 4. vipsAreDualStack(vips) returns true, overriding --dual-stack=false
 5. nodeip-configuration waits for an IPv6 address that is actively being removed from the node

cmd/runtimecfg/node-ip.go
```diff
@@ -294,6 +325,54 @@ func parseIPs(args []string) ([]net.IP, error) {
 	return ips, nil
 }
 
+// vipsAreDualStack returns true if the given VIPs contain both IPv4 and IPv6
+// addresses, indicating that the cluster is configured as dual-stack.
+func vipsAreDualStack(vips []net.IP) bool {
```
rbbratta: https://github.com/openshift/machine-config-operator/pull/6395   filter on-prem nodeip VIP args by IPFamilies

wild idea to filter the VIPs based on IP family first.

doesn't help in the event of un-authorize multiple NODEIP_HINTS case, but we will eventually fix that by adding shell quoting.

cmd/runtimecfg/node-ip.go
```diff
@@ -228,20 +266,45 @@ func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType stri
...
 						return nil, false, fmt.Errorf("Failed to find node IP")
 					}
-					time.Sleep(time.Second)
+					d.sleep(time.Second)
+					continue
+				}
+				if needsDualStackWait(dualStack, retry, chosen) {
+					logDualStackWait(&dualStackWaitStart, &lastHeartbeat, chosen)
+					d.sleep(time.Second)
 					continue
```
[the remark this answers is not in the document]
rbbratta: No retry is correct. We do not implement half stack. We only implement dual stack.  This tracks NetworkManager `ipv4.may-fail=false`  `ipv6.may-fail=false`  behavior.

cmd/runtimecfg/node-ip.go
```diff
@@ -209,17 +211,53 @@ func checkAddressUsable(chosen []net.IP) (err error) {
...
+// configuration.
+//
+// LIMITATION: split-NIC dual-stack node IPs are NOT supported. On the VIP path
+// the second family is only discovered when it shares the SAME interface as the
+// VIP-matching address (see utils.AddressesRouting). If a node's IPv4 and IPv6
+// node IPs live on different interfaces, only one family is found and the node
+// will wait for the second family indefinitely. Dual-stack node IPs are
+// expected to share a single interface (typically br-ex).
+func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType string, dualStack bool) (chosen []net.IP, matchesVips bool, err error) {
+	return getSuitableIPsInternal(retry, vips, preferIPv6, networkType, dualStack, ipDiscoverer{
```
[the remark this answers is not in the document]
rbbratta: Dual stack is only configured by ServiceNetwork  IPFamilies in the template. It is not configured by the presence of multiple VIPs.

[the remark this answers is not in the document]
rbbratta: see also https://github.com/openshift/machine-config-operator/pull/6395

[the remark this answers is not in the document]
rbbratta: see https://gist.github.com/rbbratta/30c069e1f6d6daa3144b731dc0914d12

cmd/runtimecfg/node-ip.go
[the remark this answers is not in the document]
rbbratta: see https://gist.github.com/rbbratta/30c069e1f6d6daa3144b731dc0914d12

cmd/runtimecfg/node-ip.go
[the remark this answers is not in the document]
rbbratta: Can't argue with the bot? :man_shrugging:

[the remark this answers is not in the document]
mkowalski: Yeah copilot is stupid and `isn't available for cross-repository pull requests `. We should not use it because our corporation does not like it. I have it so I trigger it on my work anyway. But for cross-fork requests it will not engage into any conversation

cmd/runtimecfg/node-ip.go
```diff
@@ -228,20 +289,45 @@ func getSuitableIPs(retry bool, vips []net.IP, preferIPv6 bool, networkType stri
 		utils.SetDebugLogLevel()
 	}
 
+	// The dual-stack wait only takes effect together with retry (see
+	// needsDualStackWait). Warn loudly if --dual-stack was requested without
+	// retry so the flag cannot silently do nothing and write a partial config.
+	if dualStack && !retry {
```
cybertron, line 301: Is this a valid configuration? Maybe instead of warning we should raise an error.

cmd/runtimecfg/node-ip.go
```diff
@@ -205,21 +207,80 @@ func writeToFile(path string, data string) error {
...
-func checkAddressUsable(chosen []net.IP) (err error) {
-	// If using IPv6, verify that the choosen address isn't tentative
-	// i.e. we can actually bind to it
-	if len(chosen) > 0 && net.IPv6len == len(chosen[0]) {
-		_, err = net.Listen("tcp", "["+chosen[0].String()+"]:")
-		if err != nil {
-			log.Errorf("Chosen node IP is not usable")
+// checkAddressUsable verifies that every chosen IPv6 address is usable, i.e. it
+// is not tentative and we can actually bind to it. IPv4 addresses have no
+// tentative state, so they are skipped. On a dual-stack node IPv4 is preferred
```
cybertron: I don't love all the comments that assume v4-primary is the only way to do dual stack. I'm not even sure that will be the most common case moving forward since a lot of people want to eventually move to v6 single stack.

I assume this is an AI blind spot. Can we prompt it to update this to avoid confusion for anyone who has to read this code in the future?

To be explicit: There is no "default" preference in a dual stack cluster. The preferred IP family is solely determined by the order of the networks defined at install-time. In runtimecfg we should never make assumptions about which is primary.

cmd/runtimecfg/node-ip.go
```diff
@@ -205,21 +207,80 @@ func writeToFile(path string, data string) error {
...
+// (dualStack) with retry enabled it blocks until BOTH address families are
+// available, so nodeip-configuration never writes a single-family (partial)
+// configuration.
+//
+// LIMITATION: split-NIC dual-stack node IPs are NOT supported. On the VIP path
+// the second family is only discovered when it shares the SAME interface as the
+// VIP-matching address (see utils.AddressesRouting). If a node's IPv4 and IPv6
+// node IPs live on different interfaces, only one family is found and the node
+// will wait for the second family indefinitely. Dual-stack node IPs are
+// expected to share a single interface (typically br-ex).
```
cybertron: Nit: It's typically _not_ br-ex today. Unless you manage br-ex with NMState, br-ex won't exist when nodeip-configuration runs which means it will always be on some other interface. It doesn't get bridged by configure-ovs until _after_ nodeip-configuration runs.

cmd/runtimecfg/node-ip.go
```diff
@@ -205,21 +207,80 @@ func writeToFile(path string, data string) error {
 	return nil
 }
 
-func checkAddressUsable(chosen []net.IP) (err error) {
-	// If using IPv6, verify that the choosen address isn't tentative
-	// i.e. we can actually bind to it
-	if len(chosen) > 0 && net.IPv6len == len(chosen[0]) {
```
cybertron, line 211: This is just a note for posterity, in case we ever come back to this change for some reason.

I assume from the fact that we hard-coded chosen[0] here and it didn't cause any problems in v4-primary dual stack clusters (where the v6 address would be chosen[1]) that we only cared about the primary IP being usable and only needed to check in v6-primary dual stack or v6 single stack clusters.

I see this patch checks all addresses in chosen, which I _think_ is fine, but I want to call out the fact that it's a change in behavior from what we used to do, and further that I think it's irrelevant to fixing the behavior of the v4 ACD interaction since there are no v4 tentative addresses and thus no need to check them for usability.

I wanted to note this in part so that if this change does somehow break something, we could remove the new logic for iterating chosen and leave the other --dual-stack behavior in place, since I think that's what is actually needed to fix the bug here.

cybertron (commented):
Nothing critical, but a few comments inline.

cmd/runtimecfg/node-ip.go
mkowalski: Done, https://github.com/openshift/baremetal-runtimecfg/pull/391/changes/5a41715090c09a52188b78d193754c6613ef5285. Told it to not assume IPv4 is better than IPv6 and both should be 1st class citizens

cmd/runtimecfg/node-ip.go
mkowalski: Done, dropped "typically br-ex" from this sentence

cmd/runtimecfg/node-ip.go
mkowalski: No, dual-stack without retry is not correct. I am just not sure if error **here** is the best place. But maybe we should just error here and force us/others to fix all the corner cases which do something hacky

cmd/runtimecfg/node-ip.go
mkowalski: Ok, back to checking only `chosen[0]`. I would like us to be hard on this validation in principle but I agree this PR is about fixing one particular thing and this validation is not needed for that.

PR #397, opened 2026-08-13
OCPBUGS-109633: Resign VIPs when the node shuts down

mkowalski:
Durable fix for [OCPBUGS-109633](https://issues.redhat.com/browse/OCPBUGS-109633), follow-up 6 of [openshift/machine-config-operator#6402](https://github.com/openshift/machine-config-operator/pull/6402): move the shutdown VIP-resign logic to where VRRP state and VIP knowledge live.

## Problem

During `systemctl reboot` systemd tears down all container scopes in parallel. Whether keepalived manages to send its VRRP priority-0 resign advert and remove the VIPs before being killed is a race. When it loses, the stale VIP keeps attracting new client connections which land **directly** on the local kube-apiserver — still serving its ~70s graceful drain with `readyz=false` — because the haproxy `VIP:6443→:9445` nftables redirect is torn down at the same time. Caught by Component Readiness as a regression of `[Monitor:audit-log-analyzer] API LBs follow /readyz ...` on `periodic-ci-openshift-release-main-ci-5.0-e2e-vsphere-ovn-upgrade` (regression 46508, triage 722): peers took MASTER via the 3–4s master-down timeout with no priority-0 advert received, while flagged requests kept terminating on the rebooting node's draining apiserver for ~70s.

## What this PR does

On SIGTERM, `dynkeepalived` (keepalived-monitor) now runs a shutdown resign sequence:

1. **Guard**: only acts when the *host* is shutting down — checked via the host's systemd (`chroot /host systemctl is-system-running` == `stopping`; the container already mounts `/host` and has `CAP_SYS_CHROOT`). An ordinary pod restart (MachineConfig rollout, liveness restart, upgrade) never touches the VIPs of a healthy node. Note `is-system-running` exits non-zero for every state except `running`, so the output is inspected, not the exit code.
2. **No-op** if none of the API/ingress VIPs is configured locally (netlink address comparison, so IPv6 spelling differences don't matter).
3. **Clean path**: sends `stop\n` over the existing keepalived control socket (same mechanism the bootstrap flow already uses), so keepalived sends the priority-0 advert (instant peer failover) and removes the VIPs itself; waits up to 10s.
4. **Fallback**: force-removes the VIPs via netlink. Because a still-running keepalived in MASTER state re-adds monitored addresses, removal is retried until the VIPs stay gone, bounded to 30s (< the pod's 65s termination grace).

## Notes for reviewers

- The MCO keepalived wrapper currently implements only `reload`; the bootstrap wrapper already implements `stop`. Until MCO adds `stop` support (trivial follow-up in the wrapper's `msg_handler`), masters/workers take the netlink force-removal path, which still guarantees no stale VIP survives into the drain window (peers fail over via the ~3–4s master-down timeout).
- Complements (and can eventually replace) the systemd-unit approach in openshift/machine-config-operator#6402; the two are independent and safe to run together — both paths are idempotent no-ops when the VIP is already gone.
- Timing windows are package `var`s so the unit tests can shrink them; tests cover: healthy-node restart (no-op), no local VIP (no-op), clean resign after `stop`, force-removal, keepalived fighting re-adds, and netlink listing failure.

## Testing

- `go test ./...` — new ginkgo specs in `pkg/monitor/shutdownresign_linux_test.go` pass along with the existing suite.
- `go build ./...` on linux and a `GOOS=darwin` cross-compile (non-linux stub).
- Suggested cluster verification: on a VIP-holding master, loop `systemctl reboot` ~20x; confirm the VIP disappears from the rebooting node before the apiserver drain window, peers log either `Backup received priority 0 advertisement` or take over within ~4s, and that a plain `oc delete pod`/static-pod restart of keepalived-monitor does *not* move the VIP.

---
This PR was generated using AI. Please verify before acting on it.
Assisted-By: Claude Fable 5

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved host shutdown handling by cleanly stopping keepalived and removing configured virtual IPs.
  * Added a fallback to force-remove virtual IPs that remain assigned during shutdown.
  * Prevented cleanup from running during pod-only restarts.
  * Added compatible behavior for non-Linux platforms without shutdown cleanup.

* **Documentation**
  * Documented shutdown behavior and virtual IP cleanup.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,6 +116,7 @@ dynkeepalived path_to_kubeconfig path_to_keepalived_cfg_template path_to_config
 - Manages Virtual IP configurations for API and Ingress
 - Supports different control plane topologies
 - Handles multiple VIPs
+- On SIGTERM, if the host itself is shutting down (not just the pod restarting), makes sure no VIP stays configured on the node: requests a clean keepalived stop and force-removes the VIPs as a fallback, so client traffic cannot be routed to the local draining kube-apiserver (OCPBUGS-109633)
 
 ---
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,7 +116,7 @@ dynkeepalived path_to_kubeconfig path_to_keepalived_cfg_template path_to_config
 - Manages Virtual IP configurations for API and Ingress
 - Supports different control plane topologies
 - Handles multiple VIPs
-- On SIGTERM, if the host itself is shutting down (not just the pod restarting), makes sure no VIP stays configured on the node: requests a clean keepalived stop and force-removes the VIPs as a fallback, so client traffic cannot be routed to the local draining kube-apiserver (OCPBUGS-109633)
+- On host shutdown (not mere pod restart), ensures no VIP stays configured: requests a clean keepalived stop, force-removes the VIPs as a fallback (OCPBUGS-109633)
 
 ---
 
--- a/pkg/monitor/dynkeepalived.go
+++ b/pkg/monitor/dynkeepalived.go
@@ -331,6 +331,10 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
 	for {
 		select {
 		case <-ctx.Done():
+			// The node may be rebooting/shutting down: make sure no
+			// stale VIP survives into the shutdown window
+			// (OCPBUGS-109633). No-op if only the pod is restarting.
+			handleShutdownResign(conn, append(append([]net.IP{}, apiVips...), ingressVips...))
 			return nil
 
 		case APIStateChanged := <-bootstrapStopKeepalived:
--- a/pkg/monitor/dynkeepalived.go
+++ b/pkg/monitor/dynkeepalived.go
@@ -328,13 +328,16 @@ func KeepalivedWatch(kubeconfigPath, clusterConfigPath, templatePath, cfgPath st
 		return err
 	}
 	defer conn.Close()
+	// The node may be rebooting/shutting down when this function exits -
+	// whether via SIGTERM (ctx.Done) or an error return, e.g. the draining
+	// apiserver failing a config fetch. Make sure no stale VIP survives
+	// into the shutdown window (OCPBUGS-109633). This is a no-op unless the
+	// host itself is shutting down, so ordinary pod restarts and runtime
+	// errors on a healthy node never touch the VIPs.
+	defer handleShutdownResign(conn, append(append([]net.IP{}, apiVips...), ingressVips...))
 	for {
 		select {
 		case <-ctx.Done():
-			// The node may be rebooting/shutting down: make sure no
-			// stale VIP survives into the shutdown window
-			// (OCPBUGS-109633). No-op if only the pod is restarting.
-			handleShutdownResign(conn, append(append([]net.IP{}, apiVips...), ingressVips...))
 			return nil
 
 		case APIStateChanged := <-bootstrapStopKeepalived:
--- a/pkg/monitor/shutdownresign_linux.go
+++ b/pkg/monitor/shutdownresign_linux.go
@@ -3,6 +3,7 @@
 package monitor
 
 import (
+	"errors"
 	"io"
 	"net"
 	"os/exec"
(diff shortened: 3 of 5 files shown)


mkowalski:
## Deep-review panel results (self-review before human review)

Ran a multi-specialist review (bugs / adversarial / security / architecture / consistency / QA / writer) against the initial commit. Findings and how they were addressed in e63699e:

### Blocking — fixed

1. **Error/success conflation in the resign paths** (bugs + adversarial + QA, independently corroborated). Three false-success traces existed:
   - a netlink list error at entry → logged "no VIP configured, nothing to do" and aborted the resign entirely;
   - a list error during the stop-wait → logged "all VIPs resigned cleanly" and returned;
   - `forceRemoveVIPs` returned the same `false` for "nothing left" / "list failed" / "AddrDel failed", so **3 consecutive AddrDel failures (e.g. EPERM) made the fight loop log "VIPs stayed removed" with the VIP still configured** — the exact condition this PR exists to prevent.

   Fix: list errors propagate; during a confirmed host shutdown, unknown state fails *toward resigning* (stop is still sent, removal still attempted), never toward keeping a possibly stale VIP; the fight loop counts a quiet poll only when a successful list found nothing; `EADDRNOTAVAIL`/`ENOENT` from a concurrent keepalived removal counts as removed.

2. **Resign bypassed on error-return exits** (bugs). The resign only ran on the `ctx.Done()` select case, but during host shutdown `KeepalivedWatch` is at least as likely to exit via `return err` from `config.GetConfig` against the draining apiserver — silently skipping the fix in its primary scenario. Fix: the resign is now a **deferred call** right after the control-socket dial, covering every exit path; the host-shutting-down guard keeps it a no-op on ordinary errors/restarts.

### Suggestions — implemented

- Pinned `/usr/sbin/chroot` (verified present in the RHEL9 base image) so a PATH change cannot silently disable the host-shutdown check (security).
- Undeterminable host state now logs at **Warn**, so silent deactivation of the fix is visible in must-gathers (adversarial).
- Extracted `hostStateIsStopping(out, err)` and table-tested it — including the subtlety that `systemctl is-system-running` exits non-zero for every state except `running` (QA).
- New tests: persistent AddrDel failure must *not* claim success; concurrent-removal race (EADDRNOTAVAIL); initial list failure proceeds with the resign; `conn.Write` failure falls back to force-removal; quiet-poll early exit asserted via timing bound (QA/adversarial).
- Test fakes gained error injection and a `resignAfterLists` mode instead of fragile inline stub overrides; the re-add fake's comment now matches its behavior (QA).
- AGENTS.md bullet trimmed to match sibling entries; grace-period comments now cite their sources (65s = keepalived pod `deletionGracePeriodSeconds` in MCO, ~70s = apiserver shutdown-delay window) (consistency/writer).

### Noted, intentionally not done here

- **Reuse `pkg/utils.getAddrs`** for the netlink walk: would require exporting it; kept the small local copy to avoid widening the utils API in this PR (architecture).
- **Stale control-socket conn**: the long-lived conn could point at a restarted socat peer; the write then succeeds into a dead buffer and the 10s wait is wasted — recovered by the force-removal fallback. Same latent behavior as the pre-existing reload path; a re-dial would fix both and fits a separate cleanup (bugs).
- **Slow reaction inside the select loop's default branch** (API calls before `ctx.Done()` is observed) — mitigated by the deferred resign on error exits; making the config fetches context-aware is a larger refactor (bugs).
- **MCO wrapper `stop` support** remains the follow-up that turns the clean path on for masters/workers; until then the 10s stop-wait is a dead wait there (adversarial noted this costs 10s of the 65s grace budget — still within bounds: 10s + 30s fight < 65s).

`go build` (linux + darwin cross-compile) and `go test -race ./...` pass.

---
This comment was generated using AI. Please verify before acting on it.
Assisted-By: Claude Fable 5

mkowalski:
Addressed the CodeRabbit concern about the unbounded host-state detection in 55f78d5:

- `systemctl is-system-running` (via chroot) is now bounded by a 5s context timeout, plus `Cmd.WaitDelay=1s` so an inherited stdout pipe cannot prolong the wait after the process is killed.
- On timeout the state is treated as unknown → the resign is skipped (fail-closed toward not flapping VIPs on a possibly-healthy node); keepalived's own SIGTERM handling and the MCO shutdown unit (openshift/machine-config-operator#6402) remain as backstops for that corner.
- Worst-case budget for the whole deferred resign is now provably bounded: 5s state check + 10s stop-wait + 30s fight window = 45s < the pod's 65s `deletionGracePeriodSeconds`, so pod shutdown can never be held hostage by a wedged systemd/dbus.

Rationale for fail-closed on timeout rather than fail-open: a hung systemctl can also happen on a healthy node (dbus restart, cgroup pressure), and force-removing VIPs there would cause an unnecessary cluster-wide API VIP failover — a worse outcome than occasionally deferring to the backstops during an actual shutdown.

This comment was generated using AI. Please verify before acting on it.
Assisted-By: Claude Fable 5

PR #404, opened 2026-09-07
[release-4.22] OCPBUGS-101879: Fix CVE-2026-33814 by adding patch for golang.org/x/net package

namansharma18899:
### 1. Why is this pull request needed and what does it do?
This PR Addresses the fix for CVE-2026-33814 without bumping the go version. The CVE is exploited when processing HTTP/2 SETTINGS frames, transport will enter an infinite loop of writing CONTINUATION frames if it receives a SETTINGS_MAX_FRAME_SIZE with a value of 0.

### 2. Which issues (if any) are related?
CVE-2026-33814

### 3. Which documentation changes (if any) need to be made?
None.

### 4. Does this introduce a backward incompatible change or deprecation?
No.


cybertron:
Unlike previous HTTP2 bugs, it appears this one does affect us because it's actually a DoS against an HTTP client, not a server (which we don't run in this project).

However, the bug needs to be fixed on 5.0 and 5.1 first and linked to the referenced bug in order for this to merge.

